Issue #262, opened 2026-04-17
Local API Phase D: agent onboarding + operator-mode dashboard


krisztiankoos:
Follow-up to #258 / #259 / #260 (Phase A+B+C shipped). Phase D turns the local API into a true **control plane** for humans and agents, and onboards every agent to use it.

Scope combines two requests:
1. **Agent onboarding** (user): update CLAUDE.md + GEMINI.md + AGENTS.md (Codex's rules file) so every agent treats `/api/briefing/session` + `/api/schema` as the first call, understands the ETag/304 convention, and knows how to use the Phase C endpoints (leases before claiming a module, diagnostics before fixing, reviews before re-reviewing).
2. **Operator-mode dashboard** (Codex's review of what Phase C doesn't yet cover for the humans who actually run this thing).

## Codex's wishlist (verbatim, ranked by impact)

> **A clear "operator mode" homepage: Now, Blocked, Next.**
>  - **Now** — what's in flight right this second: active leases, running workers, in-progress modules, unread inbox messages.
>  - **Blocked** — stuck leases, rejected modules, PRs awaiting review, issues waiting on a human, budget-capped sections.
>  - **Next** — highest-priority modules in the queue, ready to pick up; sections approaching "production ready" but not quite there; PR's one approval away from merge.

> **Section-level production readiness, not just global counts.** Per-track, per-section (Foundations, Disciplines, Toolkits, CKA Part 1..5, etc.): cleared / total / in-flight / not-yet-enqueued + % readiness score.

> **Module drill-down by key with latest v2/translation/lab state in one view.** Single page that merges `/api/module/{key}/state` + `/state/orchestration/latest` + `/lease` + `/reviews?module=...` + `/state/diagnostics` so an agent/operator gets the full picture in one fetch.

> **Explicit backlog classes**:
> - Approval debt (PRs/reviews waiting)
> - Upgrade debt (outdated deps, EOL versions, deprecated APIs)
> - Curriculum debt (rubric-critical, rubric-poor, missing labs, missing UK translations)

> **Recent activity feed:**
> - Commits
> - Queue transitions (pipeline_v2 events grouped)
> - Watched issue changes
> - Runs of `dispatch.py`/`ab`

> **Stale/idle detection:**
> - No queue movement in N minutes
> - Worker alive but no state change (we have stuck_in_state — extend with worker-heartbeat view)

> **Better track rollups**: cleared / total / in-flight / not-yet-enqueued, per track and per section.

[...]

PR #268, opened 2026-04-17
Local API Phase D (part 1): activity feed, tracks readiness, stale-worker view

krisztiankoos:
First of 3 mini-PRs for #262. Backend endpoints only — no dashboard or doc changes. Keeps the review surface small so Codex can focus.

## Summary

- **`GET /api/activity?since=<epoch|ISO>&limit=`** — merged chronological feed (git commits + pipeline v2 events + bridge messages). Newest-first. Per-source caps (`4 × limit`, min 50) so a quiet source isn't crowded out at the merge boundary. Items with an unparseable timestamp are dropped rather than anchored at epoch-0.
- **`GET /api/tracks/readiness`** — per-track, per-section production-readiness grid. Every English module on disk is bucketed into one of `cleared` / `in_flight` / `dead_letter` / `not_yet_enqueued` based on the latest v2 job row for that module. Unknown `queue_state` values default to `in_flight` so a schema drift can't silently mark work cleared.
- **`GET /api/pipeline/v2/stuck`** — adds `stale_workers[]` roll-up: grouped by `leased_by`, listing workers holding non-expired leases that are silent across ALL of them. Catches zombies the per-module `stuck_in_state` view can miss when a sibling lease fires events. Briefing gains one extra alert line for stale workers.
- `/api/schema` lists all three new / extended endpoints.

## Scope cuts (deferred)

Held for follow-up PRs on the same #262 tracker:

- Dashboard Operator tab (Now / Blocked / Next columns driven by `actions.*`, per-section grid, activity widget).
- Agent-onboarding doc updates (`GEMINI.md`, `AGENTS.md`, `gemini-workflow.md`, `scripts/agent_onboarding.md`).
- `/api/operator/now|blocked|next` and `/api/workers/heartbeat` — the handoff flagged these as redundant with the already-shipped `actions.*` and the new `stale_workers[]`; leaving them out unless a concrete gap appears.

## Tests

18 new tests:

- activity feed: empty repo, commit + event merge-sort, `since` filter, ISO `since`, bad `since` → 400, `_iso_to_epoch` round-trip (Z and `+00:00`), schema list.
- tracks readiness: empty repo, cleared-vs-not-yet-enqueued ratio, latest-state-wins for repeated module rows, dead-letter separated from in-flight, top-level modules use `_root`, schema list.
- stale workers: multi-lease grouping, single-lease heartbeat keeps worker alive, expired leases excluded, `None` idle sorts before finite idle.

`pytest tests/test_local_api.py -q` → **86 passed**. `ruff check` → clean.

## Bench

`scripts/bench_orientation.py` — worktree numbers:

| mode        | tokens | vs baseline |
|-------------|-------:|:------------|
| baseline    | 4,463  | —           |
| full        | 1,513  | **66 %**    |
| compact     | 1,069  | **76 %**    |

Compact still well above the 60 % Phase D target.

## Test plan

- [ ] Codex adversarial review → APPROVE (will route via `scripts/ab ask-codex`)
- [ ] `curl /api/activity?limit=5 | jq` on a live API returns a merged stream
- [ ] `curl /api/tracks/readiness | jq '.totals'` returns non-zero `total` against the primary repo
- [ ] `curl /api/pipeline/v2/stuck | jq '.stale_workers_count'` is an integer

Ref #262.

### DIFF
--- a/.claude/rules/gemini-workflow.md
+++ b/.claude/rules/gemini-workflow.md
@@ -4,6 +4,18 @@ description: How to work with Gemini via direct CLI dispatch
 
 # Gemini Multi-Agent Workflow
 
+## Agent Orientation (first call)
+
+Before drafting or reviewing, pull the local-API briefing so Gemini's context is aligned with what's actually in flight:
+
+```bash
+curl -s http://127.0.0.1:8768/api/briefing/session?compact=1
+curl -s http://127.0.0.1:8768/api/module/{module_key}/state   # diagnostics[] before fixing
+curl -s http://127.0.0.1:8768/api/reviews?module={key}        # existing review log before re-reviewing
+```
+
+If the API is down, fall back to `STATUS.md` + `git log -20`. See [`scripts/agent_onboarding.md`](../../scripts/agent_onboarding.md) for the full recipe including lease checks and readiness/activity feeds.
+
 ## Dispatch Command
 ```bash
 # Simple review (stdout)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,22 @@
 # Repository Guidelines
 
+## Agent Orientation (first call on a cold start)
+
+**Do not start with `git log` or a grep sweep.** Hit the local API first:
+
+```bash
+curl -s http://127.0.0.1:8768/api/briefing/session?compact=1   # ~0.7K tokens
+curl -s http://127.0.0.1:8768/api/schema                       # endpoint index
+```
+
+The briefing returns `actions.{active, blocked, next}` + `top_modules[]` — answers "what should I touch" in the same call as "what is the global state". Responses carry a weak ETag; send `If-None-Match` for 304 on repeat polls.
+
+Before claiming work: `GET /api/pipeline/leases`. Before fixing a module: `GET /api/module/{key}/state` (structured `diagnostics[]`). Before re-reviewing: `GET /api/reviews?module={key}`. Situational awareness: `GET /api/tracks/readiness` + `GET /api/activity`.
+
+Full recipe: [`scripts/agent_onboarding.md`](scripts/agent_onboarding.md). If the API is down, fall back to `STATUS.md` + `CLAUDE.md`.
+
+Everything below this block is historical MkDocs-era guidance; for current site structure see the Curriculum Structure section in `CLAUDE.md`.
+
 ## Project Structure & Module Organization
 - Course content lives under `docs/` and is grouped by track: `docs/prerequisites/`, `docs/linux/`, `docs/k8s/`, and `docs/platform/`. Each track contains numbered `module-X.Y-name.md` files plus local `README.md` overviews.
 - Navigation is defined in `mkdocs.yml`; add new pages there to surface them in the site.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -12,7 +12,14 @@ curl -s http://127.0.0.1:8768/api/briefing/session?compact=1   # ~0.7K tokens, c
 curl -s http://127.0.0.1:8768/api/schema                       # endpoint index
 ```
 
-The briefing covers: current branch + dirty summary, all worktrees, runtime services, pipeline v2 queue head, recent commits, top TODO bullets, blockers, and alerts. Drill-down URLs are listed in `next_reads`. Responses carry a weak ETag — send `If-None-Match` for 304 on repeat polls. If the API is down, fall back to reading `STATUS.md` + `CLAUDE.md`.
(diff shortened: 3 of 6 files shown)


krisztiankoos:
## Codex adversarial review — round 1

**Verdict: NEEDS CHANGES** (3 real bugs + 1 comment)

1. `build_tracks_readiness` misclassifies real pipeline state. The code expected pseudo-states like `done`, `pending_review`, and `dead_letter`, but the live v2 DB only stores `pending | leased | completed | failed` in `jobs.queue_state` (`scripts/pipeline_v2/control_plane.py:195-201`). Dead-letter is derived from events in the reducer used by `_build_status_report` (`scripts/pipeline_v2/cli.py:458-560`), so unresolved dead letters would show up as `in_flight`. Tests masked this by seeding impossible rows like `queue_state='dead_letter'`.
2. `/api/activity` can return the wrong "newest-first" pipeline items. It pulled candidates from `build_pipeline_events()`, which limits by `ORDER BY id DESC` — not by `at`. A high-id backfilled event with an old timestamp could consume one of the `4×limit` slots and push a genuinely newer event out of the candidate set.
3. Bridge-row filtering used a lexical string compare. `_iso_to_epoch()` accepts fractional-second and `+00:00` timestamps, but the SQL filter used `WHERE timestamp >= ?` with a `…Z` cutoff — lexically, `…Z` sorts after `…0.500Z` and `…+00:00`, so valid newer rows were dropped.

Comments:
- `stale_workers` tie ordering is not deterministic (no `ORDER BY` in source query, no tiebreaker in final sort).
- `MAX(id)` itself is fine given `jobs.id` is `AUTOINCREMENT`.
- No `CACHE_POLICY` entries for the new endpoints is okay for `/api/activity`; for `/api/tracks/readiness` it's a perf choice, not a correctness problem.

Reviewed by source inspection only; did not execute the Phase D branch tests.

*Posted by Claude on behalf of Codex (sandbox lacks GitHub write).*

---

## Codex adversarial review — round 2 (commit `94d91f2b`)

**Verdict: No findings.**

Re-reviewed against the three prior blockers plus the stale-worker ordering comment. The fixes look correct: `/api/tracks/readiness` now buckets off the same reducer inputs as `/api/pipeline/v2/status`, `/api/activity` selects pipeline candidates by `at DESC` instead of `id DESC`, and bridge `since` filtering now goes through `_iso_to_epoch` in Python so fractional/offset ISO rows are preserved.

Residual risk is only maintainability: `_load_v2_module_statuses` still mirrors `_build_status_report`'s job/event aggregation logic instead of sharing a single helper, so future reducer-input changes could drift again if only one side is updated. **Not a blocking issue for this patch.**

Testing gap: static review of commit `94d91f2b` from the current read-only `main` worktree; did not execute the suite from `claude/local-api-phase-d`.

*Posted by Claude on behalf of Codex.*

krisztiankoos:
## Codex adversarial review — round 3 (commits `6b26d728`, `e1053aaa`)

**Verdict: NEEDS CHANGES.**

**Bugs**

1. Medium: `renderOperator()` misroutes drill-downs when the same module appears in multiple action buckets. The JS de-duplicates `top_modules[]` by `module_key` and keeps the first endpoint it sees, then does a substring scan against the rendered row text. A module in both `active` (endpoint = `/api/module/{key}/state`) and `blocked` (endpoint = `/api/pipeline/v2/events?...`) gets the wrong link for one of the two buckets.
2. Medium: `renderOperator()` drops drill-down links for non-module action rows even though the briefing supplies endpoints for them (rows like "N job(s) ready to pick up in pipeline v2" → `/api/pipeline/v2/status`).
3. Low: `scripts/agent_onboarding.md` states the wrong timestamp contract for merged activity-feed rows. The doc says bridge-backed rows use ISO-8601 strings, but `/api/activity` normalizes bridge timestamps to Unix-epoch seconds on `items[].at`. Only `/api/bridge/messages` preserves the ISO string.

**Preferred fix**: stop reverse-parsing display strings. Have `/api/briefing/session` emit structured action rows like `{bucket, label, module_key, reason, endpoint}` and render those directly.

*Posted by Claude on behalf of Codex.*

---

## Codex adversarial review — round 4 (commit `5a0c8d94`)

**Verdict: APPROVED.**

`action_rows` closes both drill-down bugs cleanly: the dashboard now renders from per-row structured data instead of reverse-parsing labels, and the new tests cover both the cross-bucket misroute case and the `module_key = None` row-drop case. The onboarding doc fix also matches the actual `/api/activity` normalization behavior.

Testing gap: Codex's sandbox couldn't run the suite (`FileNotFoundError` for tmp dir) or reach the live API — this is a diff + test inspection, not an independent runtime re-verify.

*Posted by Claude on behalf of Codex.*

PR #281, opened 2026-04-18
fix(pipeline): serialize parallel worker state mutations (#235)

### LINKED ISSUE #235: Pipeline convergence: Codex architectural review + fixes
## Context

Full architectural review of `scripts/v1_pipeline.py` (4077 lines) by Codex. The pipeline processes 816 modules through WRITE → REVIEW → CHECK → SCORE but isn't converging — modules rebuild 3-4x without settling, 124 modules were stuck in dead-end states.

### CRITICAL

- **Parallel race condition** (`run-section --workers > 1`): Shares one mutable `state` dict across threads AND lets workers `git add/commit` concurrently. State snapshots become nondeterministic. (L3049-3063, L190-201, L2848-2864)

### HIGH

- **`data_conflict` blocks on hypothetical claims**: Topic-level ledger blocks before writing, so modules stall on claims the final draft would never make. (L2231-2247, L365-400)
- **`needs_independent_review` is dead code**: `STRUCTURAL_REVIEW_INDEPENDENCE_RELAXED = True` makes fallback same-family approvals settle as `done`. (L170, L2144-2153, L2545-2548)
- **Content-aware fact ledger truncates at 40k chars**: Late-section claims pass ungrounded. (L547-555, L979-995)
- **Retry policy varies by execution mode**: Serial=4 retries, parallel=2, `reset-stuck` only matches "rejected 5/6 times". Worker-mode failures strand easily. (L2126, L3049-3059, L3617-3621)
- **Rewrite retries preserve wrong baseline**: Knowledge packet extracted from damaged draft → asset loss ratchets downward. (L898, L2327)
- **`_merge_fact_ledgers()` mutates source dicts in place**: Data corruption, non-idempotent merges. (L593-607)
- **Malformed review JSON can silently approve**: Coerces invalid `checks` to `[]`, `compute_severity` returns `clean`. (L1313, L1647, L1124)
- **Stale resume metadata overrides fresh restart**: Old `plan`/`targeted_fix`/staging from previous run leaks into new run. (L2253-2264, L2277-2284)
- **Write-only doesn't preserve draft for review**: Sets `phase="write"` so next run re-writes instead of reviewing saved draft. (L2373-2375)
- **Review fallback chain aborts prematurely**: Same-family/unapproved fallback → hard failure, skips last-resort. (L2463-2476)
- **Off-by-one in retry loop**: 100% edit apply on last attempt records "rejected N times" instead of allowing re-review. (L2605-2610, L2666-2670)
- **CHECK retry is cross-run, not in-function**: `return False` increments outer circuit breakers, batch can halt after recoverable failures. (L2759-2786)
- **`check_failures` is a lifetime counter, not consecutive**: Never cleared on success, premature terminal failure. (L2760-2763)
- **CHECK failure leaves stale `targeted_fix`**: Severe rewrite routed through targeted writer model. (L2645, L2779-2782)

### MEDIUM

- **Content-aware fact ledger merged into state poisons retries**: Draft-specific claims from rejected attempt feed into subsequent writes/integrity checks. (L2399-2415)
- **`sonnet_anchor_failures` circuit breaker incoherent**: Claims consecutive count but never compares check IDs, doesn't reset on full rewrite. (L2658, L2542, L2601)
- **`apply_review_edits` rejects valid paired insertions**: `insert_before`/`insert_after` on same anchor treated as overlap. (L1511)
- **Integrity gate plan includes warnings**: Overconstrained rewrite adds churn. (L2430)
- **`count_assets` too coarse for 20% hard gate**: Context-free regexes miscount tables/details in code blocks. Small counts → "lose one and fail". (L872-881, L2093-2102)
- **Knowledge packet/asset count disagree on Ukrainian prompts**: UK modules fail 20% gate even if rewrite followed packet. (L848-850, L879-880)
- **`data_conflict` loops forever for stable upstream ambiguity**: No manual override state. (L2160-2165)
- **Append-only errors poison future runs**: Old failures cause incorrect resets and misleading status. (L2192-2245, L3530-3537)
- **`cmd_reset_stuck` multi-marker bug**: Later branch reintroduces errors earlier branch removed. (L3594-3619)
- **`cmd_reset_stuck` misses parallel-mode rejections**: Only matches "rejected 5/6 times", not "rejected 3 times". (L3617)
- **e2e circuit breaker doesn't halt**: Breaks resume loop but section pass reruns same modules. (L3762, L3768)
- **Integrity gate raw-text claim matching**: Scans inside code fences/quotes → false severe rewrites. (L1855, L2036)
- **`_atomic_write_text` temp-file name race**: PID+microsecond suffix can collide. (L1482)

### LOW

- **Several non-module writes bypass atomic-write discipline**: Partial index/report/dashboard on crash. (L1280, L2923, L3008)
- **CLI alias inconsistency**: Advertised aliases silently fall through to nonexistent paths. (L3647-3683, L4030-4033)
- **`_find_anchor` fuzzy matching unsound for YAML/code**: Whitespace collapse loses semantic indentation. (L1406)

## Top 5 Convergence Fixes (Codex recommendation)

1. **Disable threaded execution** until state persistence and git commits are serialized through a single writer
2. **Stop blocking on pre-write topic-ledger conflicts** — only block on post-write content-aware ledger
3. **Make independent re-review real** — force REVIEW before `done` if approved by writer family
4. **Chunk content-aware fact ledger across full module** — promote unmapped claims to errors once ledger exists
5. **Unify retry/reset policy** across serial/parallel modes — drive `reset-stuck` from counters/state, not error strings

## State Machine Assessment

**Targeted fixes, not a redesign.** The phase graph is serviceable. The churn comes from unsafe orchestration, policy/transition bugs, and sticky counters — not from the phase structure itself.

## Process

- [x] Codex: full review (8 batches)
- [ ] Codex: implement fixes + test
- [ ] Gemini: adversary review of Codex fixes
- [ ] Detailed report of both Codex and Gemini findings

## Files

- `scripts/v1_pipeline.py` (4077 lines)
- `scripts/checks/structural.py` (206 lines)
- `scripts/checks/ukrainian.py` (122 lines)
- `scripts/dispatch.py` (648 lines)
- `.pipeline/codex-review-brief.md` (review brief)

krisztiankoos:
## Summary
- give each parallel run-section worker an isolated per-module state snapshot instead of sharing one nested state dict
- merge worker-local module state back into the shared state file under a dedicated state lock, while serializing git add/git commit under a separate git lock
- add a threaded regression test that runs two concurrent workers and proves both module states survive repeated concurrent saves

## Why this approach
The old parallel path only serialized save_state() and git commands, but the nested module state objects were still shared across worker threads. That left ms[...] mutations racy and made persisted snapshots nondeterministic. Per-worker snapshots remove shared mutable state from the hot path, and the explicit merge keeps the on-disk state file authoritative without serializing the whole worker execution.

## Verification
- python -m py_compile scripts/v1_pipeline.py
- python -m unittest scripts.test_pipeline.TestStateManagement.test_parallel_worker_state_merges_do_not_clobber_other_modules
- python -m unittest scripts.test_pipeline (existing unrelated failure: TestStatusFourStage.test_cmd_status_prints_four_stage_completion_table only fails in full-suite order because cmd_status() output is suppressed when earlier tests leave _quiet=True; the test passes in isolation and was not changed in this PR)

Closes #235.

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -582,6 +582,85 @@ class TestStateManagement(unittest.TestCase):
         result = p.find_module_path("../../etc/passwd")
         self.assertIsNone(result)
 
+    def test_parallel_worker_state_merges_do_not_clobber_other_modules(self):
+        """Parallel run-section workers should save isolated module snapshots."""
+        import v1_pipeline as p
+
+        content_root = Path(self.tmpdir) / "docs"
+        module_a = content_root / "track" / "module-1.1-a.md"
+        module_b = content_root / "track" / "module-1.2-b.md"
+        module_a.parent.mkdir(parents=True, exist_ok=True)
+        module_a.write_text("---\ntitle: A\n---\n\nBody\n")
+        module_b.write_text("---\ntitle: B\n---\n\nBody\n")
+
+        shared_state = {"modules": {}}
+        observed_local_keys: dict[str, set[str]] = {}
+        results: dict[str, bool] = {}
+        failures: list[BaseException] = []
+        barrier = threading.Barrier(2)
+        previous_state_lock = p._PARALLEL_RUN_SECTION_STATE_LOCK
+        previous_git_lock = p._PARALLEL_RUN_SECTION_GIT_LOCK
+
+        def fake_run_module(module_path, state, max_retries=4, models=None,
+                            dry_run=False, refresh_fact_ledger=False,
+                            write_only=False):
+            key = p.module_key_from_path(module_path)
+            ms = p.get_module_state(state, key)
+            ms["errors"] = []
+            ms["phase"] = "review"
+            ms["passes"] = True
+            barrier.wait(timeout=5)
+            for counter in range(1, 6):
+                ms["counter"] = counter
+                p.save_state(state)
+            observed_local_keys[key] = set(state["modules"].keys())
+            return True
+
+        def run_worker(module_path: Path) -> None:
+            key = p.module_key_from_path(module_path)
+            try:
+                results[key] = p._run_module_in_parallel(
+                    module_path,
+                    shared_state,
+                    max_retries=2,
+                )
+            except BaseException as exc:  # pragma: no cover - surfaced below
+                failures.append(exc)
+
+        with patch.multiple(p, CONTENT_ROOT=content_root, STATE_FILE=self.state_file):
+            with patch("v1_pipeline.run_module", side_effect=fake_run_module):
+                p._PARALLEL_RUN_SECTION_STATE_LOCK = threading.Lock()
+                p._PARALLEL_RUN_SECTION_GIT_LOCK = threading.Lock()
+                try:
+                    threads = [
+                        threading.Thread(target=run_worker, args=(module_a,)),
+                        threading.Thread(target=run_worker, args=(module_b,)),
+                    ]
+                    for thread in threads:
+                        thread.start()
+                    for thread in threads:
+                        thread.join()
+                finally:
+                    p._PARALLEL_RUN_SECTION_STATE_LOCK = previous_state_lock
+                    p._PARALLEL_RUN_SECTION_GIT_LOCK = previous_git_lock
+
+        if failures:
+            raise failures[0]
+
+        key_a = "track/module-1.1-a"
+        key_b = "track/module-1.2-b"
+        self.assertEqual(observed_local_keys[key_a], {key_a})
+        self.assertEqual(observed_local_keys[key_b], {key_b})
+        self.assertTrue(results[key_a])
+        self.assertTrue(results[key_b])
+        self.assertEqual(shared_state["modules"][key_a]["counter"], 5)
+        self.assertEqual(shared_state["modules"][key_b]["counter"], 5)
+
+        loaded = yaml.safe_load(self.state_file.read_text())
+        self.assertEqual(set(loaded["modules"]), {key_a, key_b})
+        self.assertEqual(loaded["modules"][key_a]["counter"], 5)
+        self.assertEqual(loaded["modules"][key_b]["counter"], 5)
+
 
 # ---------------------------------------------------------------------------
 # Test: Review audit log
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-race, role-swap convention)

**LGTM with minor notes** — cannot formally approve since PR author and reviewer share the user's GH identity.

### What the fix gets right

- **State lock vs git lock are separate** — parallel git commits were needlessly serializing state merges. Splitting lets state merges proceed concurrently with an in-flight commit.
- **Per-worker snapshot via `_parallel_worker_state`** — deepcopy on entry, merge under state lock on save. This eliminates the "two workers mutate `state['modules'][key]` through shared references" class of bugs entirely, not just the writes.
- **threading.local() context** — `save_state()` auto-detects parallel-worker context. Downstream callers don't need to know about the parallel split, which keeps the change well-contained.
- **Test proves the regression** — `test_parallel_worker_state_merges_do_not_clobber_other_modules` uses real threads to demonstrate concurrent saves survive.

### Nitpicks (non-blocking)

1. `_PARALLEL_RUN_SECTION_CONTEXT.value = None` in the `finally` is fine, but `delattr(_PARALLEL_RUN_SECTION_CONTEXT, 'value')` would be slightly cleaner. `getattr(..., 'value', None)` still works with either.
2. If a worker crashes mid-`run_module`, the local state is **not** merged back. That's the right choice (exception-safe), but worth a docstring line on `_run_module_in_parallel` stating this so future readers don't assume partial-merge semantics.
3. Unrelated existing flake (`TestStatusFourStage` order dependency) — noted in PR body, agreed not in scope.

### Budget

+184/-28 across 2 files (v1_pipeline.py + test_pipeline.py). Within the ≤400 LOC, ≤3 files budget from the #235 delegation.

### Recommendation

**Merge after** confirming the pre-existing `test_pipeline` flake isn't amplified (running the full suite once in CI order should be enough).

This PR closes the parallel-race line item of #235. Other HIGH/MEDIUM items on #235 remain open as separate work.

PR #284, opened 2026-04-18
fix(pipeline): defer data_conflict check until after draft

krisztiankoos:
## Summary
- stop blocking on topic-level fact-ledger conflicts before WRITE runs
- gate `data_conflict` only after the content-aware ledger inspects the generated draft
- preserve the drafted content in staging when an actual drafted claim still trips the gate
- add tests for both the false-positive case and the real drafted-conflict case

## Verification
- `python -m py_compile scripts/v1_pipeline.py`
- `python -m unittest scripts.test_pipeline.TestRunModuleSplitReviewer.test_topic_conflicts_not_in_final_draft_do_not_block scripts.test_pipeline.TestRunModuleSplitReviewer.test_data_conflict_triage_exit`
- `python -m unittest scripts.test_pipeline` *(fails in existing unrelated test: `TestStatusFourStage.test_cmd_status_prints_four_stage_completion_table`)*

Refs #235

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -2981,14 +2981,62 @@ class TestRunModuleSplitReviewer(unittest.TestCase):
 
         self.assertEqual(events[:3], ["ledger", "write", "review"])
 
+    def test_topic_conflicts_not_in_final_draft_do_not_block(self):
+        import v1_pipeline as p
+
+        topic_conflict_ledger = {
+            "as_of_date": "2026-04-12",
+            "topic": "Conflict",
+            "claims": [
+                {
+                    "id": f"C{i}",
+                    "claim": f"Hypothetical claim {i}",
+                    "status": "CONFLICTING",
+                }
+                for i in range(1, 6)
+            ],
+        }
+        review_ok = {
+            "verdict": "APPROVE",
+            "checks": [{"id": cid, "passed": True} for cid in p.CHECK_IDS],
+            "edits": [],
+            "feedback": "",
+        }
+        content_ledger = sample_fact_ledger() | {"content_aware": True}
+        state = {"modules": {}}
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-data-conflict)

**LGTM** — the fix matches the #235 spec exactly and the test plan is solid.

### What the PR gets right
- **Moved the gate from pre-WRITE to post-WRITE.** The whole bug was "topic-level ledger blocks before writing, so modules stall on claims the final draft would never make." This PR flips it to "gate only after the content-aware ledger inspects the generated draft." That's the correct fix, not a workaround.
- **Preserves drafted content in staging when an actual drafted claim trips the gate.** This matters — previously a false-positive and a true-positive both discarded work. Now a true-positive keeps the draft so a human can see what the conflict is against.
- **Two tests, symmetric coverage:**
  - `test_topic_conflicts_not_in_final_draft_do_not_block` — proves the false-positive class (the exact bug in #235)
  - `test_data_conflict_triage_exit` — proves a real drafted conflict still exits correctly

### Nitpicks (non-blocking)
1. Consider a third test: what if the draft is partial (WRITE crashed mid-generation) — does the gate see the partial draft or fall back to the topic ledger? Probably fine either way, but worth an explicit assertion so future readers know.
2. `Refs #235` in the body is good; worth also noting this closes the specific HIGH-severity line item rather than all of #235.

### Budget
+83/-21 across 2 files. Well within the ≤400 LOC / ≤3 file budget.

### Recommendation
**Merge.** Pre-existing `TestStatusFourStage` flake correctly scoped out.

This closes one of the 14+ HIGH/MEDIUM bugs from #235. The top CRITICAL (parallel race) is PR #281. Remaining items still need per-bug delegation.

---

PR #285, opened 2026-04-18
fix(pipeline): malformed review JSON fails closed, never silently approves

krisztiankoos:
## Summary
- fail closed when review payloads are malformed or omit a valid `checks` list
- route `APPROVE` through computed payload severity so invalid reviews cannot silently pass
- add regressions for malformed review payloads and missing `checks`

## Verification
- ruff check scripts/v1_pipeline.py scripts/test_pipeline.py  # fails on pre-existing E402/F541 in these files
- python scripts/test_pipeline.py TestComputeSeverity
- python -m unittest scripts.test_pipeline.TestLegacyReviewerCompat

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -1350,6 +1350,17 @@ class TestComputeSeverity(unittest.TestCase):
         sev = self.p.compute_severity("REJECT", self.all_pass, [])
         self.assertEqual(sev, "severe")
 
+    def test_malformed_review_payload_is_severe(self):
+        """Malformed review JSON must fail closed, never silently approve."""
+        self.assertEqual(self.p.compute_review_payload_severity(None), "severe")
+
+    def test_missing_checks_key_is_severe(self):
+        """APPROVE without `checks` is invalid review output → severe."""
+        self.assertEqual(
+            self.p.compute_review_payload_severity({"verdict": "APPROVE", "edits": []}),
+            "severe",
+        )
+
     def test_reject_five_failures_is_severe(self):
         """5+ failed checks → severe, always."""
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-malformed-json)

**LGTM** — fail-closed semantics are the right call, and the regression tests lock in the behavior.

### What the PR gets right
- **Fail closed** on malformed payloads. Every review-integrity decision now goes through a computed severity, not a default-pass coercion. This is exactly the anti-pattern that #235 called out at L1124/1313/1647.
- **`APPROVE` routes through computed severity.** Before: valid-looking structure → `[]` checks → `clean`. After: if severity can't be safely computed, the review doesn't silently pass.
- **Two tests, symmetric coverage:** malformed JSON payload, and missing-`checks` key. These are precisely the two observed failure modes.

### Nitpicks (non-blocking)
1. Would be worth a third test covering the "checks is a dict, not a list" malformation. Reviewers occasionally emit `{"PRES": {...}, "FACT": {...}}` when the spec wants an array of objects. The fail-closed path should cover that too. If the current implementation already rejects non-list `checks`, a one-line assertion test documents it.
2. No error-logging on the reject path that I can see in the summary. When a review fails closed, we should probably log *why* (malformed? missing key? wrong type?) so the next reviewer retry has context. Not blocking — file as follow-up if useful.

### Budget
+53/-19 across 2 files (v1_pipeline.py + test_pipeline.py). Well within budget.

### Recommendation
**Merge.** Pre-existing ruff E402/F541 noise correctly scoped out.

Closes another #235 HIGH-severity bug. 12+ remaining.

---

PR #287, opened 2026-04-18
fix(pipeline): _merge_fact_ledgers is pure and idempotent (#235)

krisztiankoos:
## Summary
- deep-copy both ledger inputs before merging so source dicts are never mutated in place
- return deep-copied single-ledger results for the one-sided cases to avoid aliasing callers to source state
- add regression coverage for unchanged inputs, repeated-call idempotence, and copy semantics

## Verification
- python scripts/test_pipeline.py TestContentAwareFactLedger -v
- ruff check scripts/v1_pipeline.py scripts/test_pipeline.py  # existing baseline failures in these files (pre-existing E402/F541)

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -19,6 +19,7 @@ import tempfile
 import threading
 import textwrap
 import unittest
+import copy
 from argparse import Namespace
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-merge-ledgers)

**LGTM** — pure-and-idempotent is the correct semantic, and the tests cover the three most important invariants.

### What the PR gets right
- **Deep-copy both inputs before merge.** Source dicts never mutated. This is the textbook fix for "in-place mutation of arguments" bugs.
- **Deep-copy single-ledger results too.** Callers no longer alias to source state on the one-sided cases. Prevents sneaky follow-on mutations.
- **Three regressions cover the invariants:**
  - Unchanged inputs (purity)
  - Repeated-call idempotence (running merge twice = running it once)
  - Copy semantics (return is disjoint from inputs)

### Nitpicks (non-blocking)
1. Deep-copy on every call has a performance cost. If the ledger gets large (it grows with every module), this becomes O(N) on every merge. Not a concern for current scale, but a docstring note ("deep copy is load-bearing for correctness; if ledger grows large, consider a freeze/copy-on-write wrapper") would help future readers.
2. Consider a fourth test: merging `a` with an empty ledger `{}` returns a deep copy of `a` (not `a` itself). This is the "trivial one-sided" case and worth an explicit assertion.

### Budget
+52/-10 across 2 files. Well within budget.

### Recommendation

Closes another #235 HIGH-severity bug. Running count: 3 of 14+ #235 HIGH bugs closed this session (parallel race, data_conflict, malformed JSON, now merge-ledgers).

Actually, that's 4. Math is hard.

---

PR #288, opened 2026-04-18
fix: revive needs_independent_review for last-resort approvals (#235)

krisztiankoos:
## Summary
- remove the dead `STRUCTURAL_REVIEW_INDEPENDENCE_RELAXED` gate from the structural review path
- track when review succeeded only via the writer-family last-resort reviewer and persist `needs_independent_review=True` for that case
- add a regression test covering the primary rate-limit -> fallback rate-limit -> last-resort approve flow

## Verification
- `ruff check scripts/v1_pipeline.py scripts/test_pipeline.py` *(fails on pre-existing E402/F541 in these files; no new lint violations introduced by this patch)*
- `pytest -q scripts/test_pipeline.py`

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -1076,6 +1076,45 @@ class TestPipelineTransitions(unittest.TestCase):
         self.assertEqual(mock_review.call_count, 2)
 
     @patch("v1_pipeline.STATE_FILE")
+    @patch("v1_pipeline.CONTENT_ROOT")
+    @patch("subprocess.run")
+    def test_last_resort_same_family_review_sets_pending_flag(
+        self, mock_subprocess, mock_root, mock_state,
+    ):
+        """Writer-family last-resort approvals must preserve the deferred re-review flag."""
+        import v1_pipeline as p
+
+        mock_state.__class__ = type(self.state_file)
+        mock_root.resolve.return_value = Path(self.tmpdir).resolve()
+
+        state = {"modules": {}}
+        all_pass_checks = [{"id": cid, "passed": True} for cid in p.CHECK_IDS]
+        review_sequence = [
+            {"rate_limited": True},
+            {"rate_limited": True},
+            {"verdict": "APPROVE", "severity": "clean",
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-needs-indep-review)

**LGTM** — "restore not remove" is the correct call and the regression test locks in the exact scenario the issue describes.

### Why restore-not-remove is right
Codex's reasoning is sound: `needs_independent_review` is still a real state for the last-resort reviewer path (writer-family fallback when both primary and cross-family reviewers are unavailable). The dead code was specifically the `STRUCTURAL_REVIEW_INDEPENDENCE_RELAXED = True` hardcoded switch that prevented the flag from ever being set. Removing the switch (not the flag) restores the intended behavior: a review that *did* come from writer-family gets flagged for a later independent pass, instead of settling as `done`.

### What the PR gets right
- **Scalpel on the switch, not the flag.** Removes `STRUCTURAL_REVIEW_INDEPENDENCE_RELAXED`; keeps `needs_independent_review` and actually writes it when appropriate.
- **Only flagged on last-resort path.** The flag is set when primary AND cross-family fallback were unavailable and we landed on writer-family. Normal primary-reviewer approvals are not incorrectly marked.
- **Regression test covers the exact chain:** primary rate-limit → cross-family fallback rate-limit → last-resort writer-family approve → `needs_independent_review = True` persists.

### Nitpicks (non-blocking)
1. Worth a docstring on the flag explaining the lifecycle: "set by X, cleared by Y (a subsequent independent reviewer), never set on happy path." Future readers will ask.
2. The PR title is `revive needs_independent_review`. "Revive" is accurate but a bit cute — the dead part was the *enforcement*, not the flag itself. Minor wording — wouldn't ask for a retitle.

### Budget
+46/-10 across 2 files. Small, surgical.

### Recommendation

This closes another #235 HIGH bug, and importantly: the "never merge without independent adversarial review" rule now has actual teeth in the pipeline, not just in the GH PR review flow.

---

PR #289, opened 2026-04-18
fix(pipeline): content-aware ledger covers late sections (#235)

krisztiankoos:
## Summary
- replace the single 40k prefix trim in `step_content_aware_fact_ledger()` with overlapping windows that cover the full module
- merge per-window ledgers with claim de-duplication so late-section claims reach REVIEW/CHECK without duplicate overlap entries
- add a regression test for >40k modules proving a final-section claim is included in the merged ledger

## Trade-offs
- chose chunk-and-merge instead of just raising the limit or head/tail sampling
- cost: long modules now make O(N) fact-grounding calls across windows, with overlap for boundary coverage
- benefit: every section is represented, middle windows may be empty without failing the pass, and memory stays bounded by one window at a time

## Verification
- `ruff check scripts/v1_pipeline.py scripts/test_pipeline.py` (repo has pre-existing `E402`/`F541` in these files)
- `ruff check scripts/v1_pipeline.py scripts/test_pipeline.py --ignore E402,F541`
- `python scripts/test_pipeline.py TestFactLedger TestContentAwareFactLedger TestEnsureFactLedger`

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -2537,6 +2537,86 @@ class TestContentAwareFactLedger(unittest.TestCase):
         finally:
             shutil.rmtree(tmpdir, ignore_errors=True)
 
+    def test_content_aware_step_covers_late_sections_beyond_40k(self):
+        import v1_pipeline as p
+
+        tmpdir = tempfile.mkdtemp()
+        try:
+            ledger_dir = Path(tmpdir) / ".pipeline" / "fact-ledgers"
+            content_root = Path(tmpdir) / "src" / "content" / "docs"
+            module_path = content_root / "test" / "module-0.1-test.md"
+            module_path.parent.mkdir(parents=True, exist_ok=True)
+            module_path.write_text("---\ntitle: Test\n---\n\nBody")
+
+            early_marker = "Kubernetes current stable is v1.35."
+            late_marker = "Helm current stable is v3.19.0."
+            filler_block = ("Padding text without factual claims.\n" * 1400)
+            content = (
+                "---\ntitle: Test\n---\n\n"
+                "## Early Facts\n"
+                f"{early_marker}\n\n"
+                f"{filler_block}\n"
+                "## Middle Narrative\n"
+                f"{filler_block}\n"
+                "## Late Facts\n"
+                f"{late_marker}\n"
+            )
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-ledger-40k)

**LGTM with one cost-concern flag** — the fix is architecturally correct, but the O(N) call count on long modules has a budget implication worth calling out.

### What the PR gets right
- **Chunk-and-merge with overlapping windows.** Every section gets represented, not just the first 40k chars. This is the correct fix for the "late-section claims pass ungrounded" bug.
- **Claim de-duplication across windows.** Overlap prevents boundary-coverage gaps but would otherwise cause duplicate entries — dedup handles that.
- **Memory bounded by one window at a time.** Doesn't balloon RAM on long modules.
- **Regression test proves it:** >40k module → final-section claim shows up in merged ledger. This is the exact failure mode #235 described.

### Cost concern (worth flagging before merge)
- **O(N) fact-grounding calls per long module.** Each window is one Codex/gpt-5.3-codex-spark call. For a 120k-char module with 40k windows and overlap, that's 3-4 calls instead of 1. Across 200+ modules this compounds: the fact-grounding budget quadruples for the large-module tail.
- PR does not discuss whether the per-window call honors the rate-limit budget (`MODELS["fact_grounding"]` concurrency cap). If it doesn't, a single long module could exhaust the hourly budget.
- Recommended follow-up: measure the N=1 vs N=4 cost over the 10 longest modules and decide if we need a module-size-aware strategy (e.g., skip grounding for windows with no code blocks or tables).

### Nitpicks (non-blocking)
1. Worth logging per-window call counts so the pipeline observability surfaces this cost immediately, not after a budget panic.
2. The "middle windows may be empty without failing the pass" rule is good but easy to regress. A test that constructs a module with ONE claim in an empty middle window would be useful.

### Budget
+205/-26 across 2 files. Largest of the #235 fixes so far — the extra lines are the windowing + merge logic, which matches the task complexity. Within ≤400 LOC budget.

### Recommendation
**Merge, but track fact-grounding call counts post-merge.** Open a follow-up issue if the cost implications bite in the next pipeline run.

Closes another #235 HIGH-severity bug. Running count: 6 of 14+ closed this session (parallel race, data_conflict, malformed JSON, merge-ledgers, needs-indep-review, ledger-40k).

---

PR #290, opened 2026-04-18
fix(pipeline): check_failures tracks consecutive, not lifetime (#235)

krisztiankoos:
## Summary
- reset check_failures to 0 on APPROVE and successful CHECK so the counter is explicitly consecutive
- add a regression test for the CHECK sequence success -> failure -> success -> failure

## Verification
- python -m unittest scripts.test_pipeline.TestPipelineTransitions.test_check_failures_tracks_consecutive_failures_only
- python scripts/test_pipeline.py (fails on existing unrelated test: TestStatusFourStage.test_cmd_status_prints_four_stage_completion_table)
- ruff check scripts/v1_pipeline.py scripts/test_pipeline.py (fails on pre-existing E402/F541 issues in these files)

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -1307,6 +1307,72 @@ class TestPipelineTransitions(unittest.TestCase):
         self.assertFalse(result)
 
 
+    @patch("v1_pipeline.STATE_FILE")
+    @patch("v1_pipeline.CONTENT_ROOT")
+    @patch("subprocess.run")
+    def test_check_failures_tracks_consecutive_failures_only(
+        self, mock_subprocess, mock_root, mock_state,
+    ):
+        import v1_pipeline as p
+
+        mock_state.__class__ = type(self.state_file)
+        mock_root.resolve.return_value = Path(self.tmpdir).resolve()
+
+        key = "test/module-0.1-test"
+        state = {
+            "modules": {
+                key: {
+                    "phase": "check",
+                    "severity": "clean",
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-check-failures-counter)

**LGTM** — one-line semantic fix with a regression test proving the oscillation pattern. Best kind of PR.

### What the PR gets right
- **Reset on both success paths.** `check_failures = 0` on APPROVE AND on successful CHECK. Prior code only reset on one path, so the "lifetime counter" behavior could still leak through. Belt-and-suspenders here is correct — a reviewer-APPROVE followed by a CHECK sequence should reset the counter, not preserve stale failures.
- **The regression test is the right shape:** success → failure → success → failure. That's exactly the oscillation pattern where a lifetime counter would trigger premature terminal failure but a consecutive counter stays at 1.

### Nitpicks (non-blocking)
1. Would also add an assertion that APPROVE resets to 0 even when coming from a prior failure — the current test covers CHECK reset but not APPROVE reset. One extra line in the test.
2. The failure path could grow a log line noting the pre-reset count ("check_failures was N before reset") for debug visibility. Optional.

### Budget
+68/-2 across 2 files. Minimal, targeted.

### Recommendation
**Merge.** Pre-existing `TestStatusFourStage` flake and ruff noise correctly scoped out.

Closes another #235 HIGH bug. Running count: 7 of 14+ closed this session.

---

PR #291, opened 2026-04-18
fix(pipeline): last-attempt edit success triggers re-review, not reject (#235)

krisztiankoos:
## Summary
- return cleanly when the last allowed attempt lands all deterministic review edits
- keep the module in `phase=review` with staged patched content for resume/re-review
- add regression tests for last-attempt deterministic-apply success vs failed-apply rejection

## Verification
- `python scripts/test_pipeline.py TestDeterministicApplyIntegration`
- `ruff check scripts/v1_pipeline.py scripts/test_pipeline.py` *(fails on pre-existing repo issues: E402/F541 in these files)*

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -3685,6 +3685,151 @@ class TestDeterministicApplyIntegration(unittest.TestCase):
         ms = state["modules"]["test/module-0.1-test"]
         self.assertEqual(ms.get("phase"), "done")
 
+    @patch("v1_pipeline.STATE_FILE")
+    @patch("v1_pipeline.CONTENT_ROOT")
+    @patch("subprocess.run")
+    def test_last_retry_full_apply_exits_in_rereview_state(
+        self, mock_subprocess, mock_root, mock_state,
+    ):
+        """A clean deterministic apply on the last retry must not reject."""
+        import v1_pipeline as p
+
+        mock_state.__class__ = type(self.state_file)
+        mock_root.resolve.return_value = Path(self.tmpdir).resolve()
+
+        review_sequence = [
+            {
+                "verdict": "REJECT",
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-off-by-one-retry)

**LGTM** — the symmetric-case test (apply-success vs apply-failure on the same last attempt) is exactly what pins down this off-by-one class of bug.

### What the PR gets right
- **Correct state transition on the boundary.** Last-attempt + edit-apply-succeeds now lands in `phase=review` with staged patched content, enabling resume/re-review. Previously it was wrongly recording "rejected N times" and terminating.
- **Staged patched content is preserved.** This matters — without the staged content, the next resume would re-run the whole WRITE→REVIEW→EDIT chain from scratch. Losing N attempts of work.
- **Symmetric tests:** apply-success transitions to review; apply-failure still rejects. Both branches of the last-attempt boundary are covered.

### Nitpicks (non-blocking)
1. Worth a docstring on the affected function (whichever contains L2605-2610) stating the "last attempt is not a rejection if edits applied" rule. Silent invariants are hard to maintain.
2. `+146` line count is larger than the fix itself suggests — likely because the test fixture setup needed to be substantial to reproduce the boundary condition. That's fine; test infra often dominates the diff for state-machine bugs.

### Budget
+146/-1 across 2 files. Within ≤400 LOC budget.

### Recommendation

Closes another #235 HIGH bug. Running count: 8 of 14+ closed this session.

### Convergence note
This fix + #281 (parallel race) + #287 (merge-ledgers pure) + #290 (consecutive counter) together should meaningfully improve convergence rate for modules that were previously flapping. The current `flapping_count: 12` metric will be the first thing to watch after all four land on main.

---

PR #293, opened 2026-04-18
fix(pipeline): fresh restart clears stale resume metadata (#235)

krisztiankoos:
## Summary
- clear stale resume-only metadata when a module is intentionally fresh-restarted
- reuse the same cleanup in both pending-start and reset-stuck paths
- add a regression test covering stale plan/targeted_fix/staging on reset-stuck

## Verification
- ruff check scripts/v1_pipeline.py scripts/test_pipeline.py --extend-ignore E402,F541
- python -m pytest scripts/test_pipeline.py -k reset_stuck

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -824,6 +824,40 @@ class TestReviewAuditLog(unittest.TestCase):
             add_paths = mock_git.call_args[0][0]
             self.assertEqual(add_paths, [str(audit_path)])
 
+    def test_reset_stuck_clears_stale_resume_metadata_on_fresh_restart(self):
+        import v1_pipeline as p
+
+        staging_path = self.module_path.with_suffix(".staging.md")
+        staging_path.write_text("stale staged draft")
+        state = {
+            "modules": {
+                self.module_key: {
+                    "phase": "check",
+                    "severity": "targeted",
+                    "checks_failed": [{"id": "LAB", "evidence": "old failure"}],
+                    "plan": "TARGETED FIX. Old plan from previous run.",
+                    "targeted_fix": True,
+                    "paused_reason": "rate limit",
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-stale-resume)

**LGTM** — `_clear_resume_metadata()` as a single cleanup function used by both entry points is the right abstraction.

### What the PR gets right
- **One cleanup function, two callers.** DRY — both the fresh-start path and `cmd_reset_stuck()` use the same `_clear_resume_metadata()`. A future third entry point (e.g., a new `--restart` CLI flag) will automatically benefit.
- **Covers the full metadata set:** `plan`, `targeted_fix`, `paused_reason`, and `.staging.md`. That matches the spec and also catches `paused_reason` which I didn't enumerate in the delegation.
- **Regression test on the reset-stuck branch.** Covers the exact "set stale metadata → trigger reset → verify cleared" scenario #235 described.

### Nitpicks (non-blocking)
1. Consider adding a test for the pending-start branch too (currently only reset-stuck is tested). The cleanup function is shared so logic is the same, but a second test locks in that BOTH entry points call it.
2. `.staging.md` cleanup is file-system side-effecting. Would be worth documenting "safe to call on modules with no staged content" (idempotent = OK). A one-line docstring on `_clear_resume_metadata()`.
3. `--extend-ignore E402,F541` is the right local workaround, but agreed with prior PRs: a repo-wide `ruff.toml` config would be cleaner.

### Budget
+50/-8 across 2 files. Tight.

### Recommendation

Running count: 10 of 14+ #235 HIGH bugs closed this session. Remaining: rewrite-retries-wrong-baseline, retry-policy-by-mode, CHECK-retry-cross-run, CHECK-stale-targeted-fix, review-fallback-chain (currently in flight).

---

PR #294, opened 2026-04-18
fix(pipeline): review fallback chain continues to last-resort instead of hard-failing (#235)

krisztiankoos:
## Summary
- treat review fallback models from the writer family as invalid and continue to the writer-model last-resort path instead of aborting
- preserve `needs_independent_review=True` when the last-resort reviewer approves so a later independent re-review is still required, matching the #288 interaction
- add regression coverage for both same-writer-family fallback skipping and approved cross-family fallback usage

## Verification
- `pytest -q scripts/test_pipeline.py -k "review_falls_back_to_sonnet_without_pending_flag or review_skips_same_writer_family_fallback_to_last_resort or review_uses_approved_cross_family_fallback_before_last_resort"`
- `pytest -q scripts/test_pipeline.py -k TestPipelineTransitions`
- `ruff check scripts/v1_pipeline.py scripts/test_pipeline.py` *(fails on pre-existing repo-wide E402/F541 lint noise in these files; no new ruff findings from this patch)*

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -1264,6 +1264,102 @@ class TestPipelineTransitions(unittest.TestCase):
         self.assertEqual(mock_review.call_count, 3)
 
     @patch("v1_pipeline.STATE_FILE")
+    @patch("v1_pipeline.CONTENT_ROOT")
+    @patch("subprocess.run")
+    def test_review_skips_same_writer_family_fallback_to_last_resort(
+        self, mock_subprocess, mock_root, mock_state,
+    ):
+        """Same-writer-family fallback should skip straight to last resort and preserve re-review flag."""
+        import v1_pipeline as p
+
+        mock_state.__class__ = type(self.state_file)
+        mock_root.resolve.return_value = Path(self.tmpdir).resolve()
+
+        state = {"modules": {}}
+        models = {
+            **p.MODELS,
+            "write": "claude-sonnet-4-6",
+            "review": "gemini-3.1-pro-preview",
+            "review_fallback": "claude-sonnet-4-6",
+        }
+        all_pass_checks = [{"id": cid, "passed": True} for cid in p.CHECK_IDS]
+        review_sequence = [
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-review-fallback-chain)

**LGTM** — treating same-writer-family fallback as invalid and continuing to last-resort is the correct semantics, and preserving `needs_independent_review=True` correctly composes with PR #288.

### What the PR gets right
- **Fallback-family check is the right choke point.** Instead of trying to re-derive "is the fallback independent", the code now just refuses to use any fallback that belongs to the writer family. Simple, auditable.
- **Last-resort path kept intact.** When the fallback is rejected, the pipeline continues to the writer-model last-resort reviewer instead of hard-failing. That's the whole point of having a last-resort in the first place.
- **Composes with #288.** The `needs_independent_review=True` flag is preserved when last-resort approves, so a downstream independent reviewer can still pass judgment. Without that preservation, we'd be "quietly approving" via writer-family which is exactly what #288 was fixing.
- **Three regression tests:**
  1. Primary falls back to Sonnet without pending flag (happy cross-family path still works)
  2. Same-writer-family fallback is skipped → last-resort invoked
  3. Approved cross-family fallback is used before last-resort (don't skip a good fallback just because you can)

### Nitpicks (non-blocking)
1. Test names are self-describing, which is great. Consider adding one more: "both primary and fallback are unavailable → last-resort still fires, `needs_independent_review=True`." The current three tests cover the decision-tree branches but not the chained-failure path explicitly.
2. If the writer-family check uses string matching on model IDs, consider centralizing to a `MODEL_FAMILIES` dict so future model additions (e.g., Gemini 4.x) don't silently bypass the check.

### Budget
+102/-3 across 2 files. Within budget.

### Recommendation
**Merge.** Pre-existing ruff noise correctly scoped out. Should be merged AFTER #288 so the interaction is intentional.

Running count: 11 of 14+ #235 HIGH bugs closed this session.

---

PR #296, opened 2026-04-18
fix(pipeline): CHECK retries in-function, shields outer circuit breaker from recoverable failures (#235)

krisztiankoos:
## Summary
- move CHECK retry handling inside `run_module` so transient deterministic-check failures do not bubble out as outer batch failures
- only persist `check_failures` and return `False` after the in-function CHECK retry budget is exhausted
- add a regression test covering fail-then-pass CHECK behavior from `phase=check`

## Verification
- `ruff check --extend-ignore E402,F541 scripts/v1_pipeline.py scripts/test_pipeline.py`
- `python -m unittest scripts.test_pipeline.TestReviewAuditLog.test_check_retries_in_function_without_returning_false`
- `python scripts/test_pipeline.py` *(currently has a pre-existing unrelated failure in `TestStatusFourStage.test_cmd_status_prints_four_stage_completion_table`)*

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -792,6 +792,40 @@ class TestReviewAuditLog(unittest.TestCase):
             self.assertIn(str(self.module_path), add_paths)
             self.assertIn(str(audit_path), add_paths)
 
+    def test_check_retries_in_function_without_returning_false(self):
+        import v1_pipeline as p
+
+        state = {
+            "modules": {
+                self.module_key: {
+                    "phase": "check",
+                    "reviewer": "gemini",
+                    "severity": "clean",
+                    "errors": [],
+                }
+            }
+        }
+        self.module_path.with_suffix(".staging.md").write_text(GOOD_MODULE)
+        transient_failure = [CheckResult("LINE_COUNT", False, "temporary check failure")]
+        git_ok = subprocess.CompletedProcess(["git"], 0, "", "")
+
+        with self._patch_paths(p), \
+             patch.object(p, "ensure_fact_ledger", return_value=sample_fact_ledger()), \
+             patch.object(
+                 p,
+                 "step_check",
+                 side_effect=[(False, transient_failure), (True, [])],
+             ) as mock_check, \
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-check-retry-in-function)

**LGTM** — composing correctly with #290 (consecutive check_failures) is the key insight, and the regression test pins down exactly the fail-then-pass behavior that was leaking out to the outer batch.

### What the PR gets right
- **Retry loop lives inside `run_module`'s CHECK phase now.** Recoverable transient CHECK failures stay local to the module run. They don't leak out as batch-level failures, don't trip the outer circuit breaker, don't halt the batch after a fix-and-retry.
- **`check_failures` only increments after in-function budget is exhausted.** This composes with #290 exactly right: #290 made `check_failures` consecutive (not lifetime), and this PR makes sure it's only incremented when the module genuinely cannot recover, not on every transient.
- **Regression test is precise:** "fail-then-pass CHECK behavior from `phase=check`" — the exact scenario where the old code would have bubbled the first-attempt failure out before the retry had a chance.

### Nitpicks (non-blocking)
1. Worth a docstring on the affected function stating: "CHECK retries are owned by this function; outer callers never see transient CHECK failures." Declares the boundary clearly.
2. The in-function retry budget is presumably a constant somewhere. Would be worth surfacing it as a configurable (env var or budgets.yaml) for operator tuning — but that's a separate task, not for this PR.
3. The pre-existing `TestStatusFourStage` flake is now mentioned in 8+ PRs this session. Worth a dedicated cleanup PR to fix the order-dependence — the flake is real signal noise at this point.

### Budget
+49/-7 across 2 files. Surgical.

### Recommendation
**Merge after #290.** The two are coupled — merging this first without #290 would leave `check_failures` as a lifetime counter that wouldn't reset on the retry-then-success path.

Running count: 13 of 14+ #235 HIGH bugs closed this session. Only `rewrite-retries-wrong-baseline` (currently in flight) and `retry-policy-varies-by-execution-mode` remain.

---

PR #295, opened 2026-04-18
fix(pipeline): severe rewrite clears stale targeted_fix, uses full rewrite model (#235)

krisztiankoos:
## Summary
- clear stale \'targeted_fix\' state before writer routing when a module resumes in severe rewrite mode
- add a regression test covering a CHECK-failure-style resume snapshot with  and stale 

## Verification
- 
- E402 Module level import not at top of file
  --> scripts/test_pipeline.py:34:1
   |
32 | sys.path.insert(0, str(REPO_ROOT / "scripts"))
33 |
34 | from checks import structural
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
35 | from checks.structural import CheckResult
   |

E402 Module level import not at top of file
  --> scripts/test_pipeline.py:35:1
   |
34 | from checks import structural
35 | from checks.structural import CheckResult
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |

E402 Module level import not at top of file
   --> scripts/v1_pipeline.py:133:1
    |
131 | sys.path.insert(0, str(REPO_ROOT / "scripts"))
132 |
133 | from checks import structural, ukrainian, gaps
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
134 | from dispatch import (
135 |     dispatch_gemini_with_retry,
    |

E402 Module level import not at top of file
   --> scripts/v1_pipeline.py:134:1
    |
133 |   from checks import structural, ukrainian, gaps
134 | / from dispatch import (
135 | |     dispatch_gemini_with_retry,
136 | |     dispatch_claude,
137 | |     dispatch_codex,
138 | |     _is_rate_limited,
139 | |     ClaudeUnavailableError,
140 | | )
    | |_^
141 |   from uk_sync import (
142 |       translate_new_module as uk_translate,
    |

E402 Module level import not at top of file
   --> scripts/v1_pipeline.py:141:1
    |
139 |       ClaudeUnavailableError,
140 |   )
141 | / from uk_sync import (
142 | |     translate_new_module as uk_translate,
143 | |     fix_module as uk_fix,
144 | |     _find_content_files as uk_find_content_files,
145 | |     CONTENT_ROOT as UK_CONTENT_ROOT,
146 | |     UK_ROOT,
147 | | )
    | |_^
    |

F541 [*] f-string without any placeholders
   --> scripts/v1_pipeline.py:974:15
    |
[...]

PR #292, opened 2026-04-18
fix(pipeline): write-only preserves draft for review, not re-write (#235)

krisztiankoos:
## Summary
- add a regression test covering the write-only draft handoff
- verify write-only leaves the saved module in phase=review
- verify the next run resumes REVIEW on the saved draft instead of invoking WRITE again

## Verification
- python -m py_compile scripts/v1_pipeline.py scripts/test_pipeline.py
- python -m pytest scripts/test_pipeline.py -q
- ruff check scripts/test_pipeline.py --extend-ignore E402
- ruff check scripts/v1_pipeline.py scripts/test_pipeline.py  # reports pre-existing repo lint debt (E402/F541), no new failures from this change

Closes #235.

PR #297, opened 2026-04-18
fix(pipeline): rewrite retries extract knowledge packet from baseline (#235)

krisztiankoos:
## Summary
- keep severe rewrite retries anchored to the committed module baseline instead of the last rejected draft
- add a regression test that simulates a damaged intermediate draft and asserts the retry knowledge packet comes from the baseline

## Verification
- `./.venv/bin/python -m py_compile scripts/v1_pipeline.py scripts/test_pipeline.py`
- `./.venv/bin/python scripts/test_pipeline.py TestPipelineTransitions.test_severity_severe_triggers_full_rewrite TestPipelineTransitions.test_rewrite_retry_uses_committed_baseline_for_knowledge_packet`
- `./.venv/bin/ruff check scripts/v1_pipeline.py scripts/test_pipeline.py` *(fails on pre-existing E402/F541 issues unrelated to #235)*

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -310,6 +310,24 @@ Continue to [Module 0.2: DNS in Kubernetes](/prerequisites/test/module-0.2-dns/)
 we explore how CoreDNS resolves Service names and how to debug DNS issues.
 """)
 
+DAMAGED_REWRITE_DRAFT = textwrap.dedent("""\
+---
+title: "Test Module — Damaged"
+slug: prerequisites/test/module-0.1-damaged
+sidebar:
+  order: 1
+---
+
+## Learning Outcomes
+
+- Recall that Kubernetes networking exists
+
+## Core Concepts
+
+This damaged rewrite dropped the code samples, tables, prompts, and quiz blocks
+that the baseline module contained.
+""")
+
 
 BAD_MODULE_NO_FRONTMATTER = "# No frontmatter\n\nSome content here."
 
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Adversarial review — Claude (via agent bridge, task-id infra-235-rewrite-baseline)

**LGTM** — anchoring rewrite retries to the committed on-disk baseline instead of the in-memory rejected draft stops the "asset loss ratchets downward" failure mode cold.

### What the PR gets right
- **Committed baseline, not rejected draft.** This is the core insight: once a retry starts with a damaged draft as input, every iteration amplifies the damage. Using the last-known-good committed version as the baseline breaks that feedback loop.
- **Regression test is the right shape:** simulates a damaged intermediate draft → triggers retry → asserts knowledge packet is derived from the baseline, not the damaged intermediate. If a future refactor inverts this, the test catches it.
- **Minimal surgery.** +106/-1 in two files. The -1 is telling: very little old code needed removal, the fix is mostly additive (the right baseline lookup + test harness).

### Nitpicks (non-blocking)
1. Test name is verbose but correct. Two variants would strengthen it:
   - "first-attempt rewrite uses committed baseline" (no prior draft to pick from, already correct)
   - "second-attempt rewrite uses committed baseline, not the first-attempt draft" (the #235 bug)
   The second covers the bug; adding the first locks in that normal-path behavior isn't accidentally changed.
2. If `committed baseline` is fetched via a `git show` or similar, consider a test fixture that stubs the git call so the test stays deterministic across machines without a full git history.
3. Worth a docstring note on the rewrite retry function: "baseline is the last committed version; the rejected draft is never used as retry input." Makes the anti-pattern explicit.

### Budget
+106/-1 across 2 files. Within ≤400 LOC budget.

### Recommendation

Running count: 14 of 14+ #235 HIGH bugs closed this session (only `retry-policy-varies-by-execution-mode` currently in flight).

### Cumulative impact
Combined with #281 (parallel race), #284 (data_conflict post-write), #285 (malformed JSON fail-closed), #287 (merge-ledgers pure), #288 (needs_independent_review), #289 (ledger 40k windowing), #290 (consecutive check_failures), #291 (last-attempt re-review), #292 (write-only regression test), #293 (fresh-restart cleanup), #294 (fallback chain last-resort), #295 (severe rewrite full-model), #296 (in-function CHECK retries), this PR (#297) rewrite-baseline: the pipeline's convergence story moves from "many modules flap" to "state mutations are deterministic and recoverable." That's the actual thesis of #235 being closed.

---

Issue #308, opened 2026-04-18
v2 pipeline: bridge review events to .pipeline/reviews/*.md audit log


krisztiankoos:
## Problem

PR #307 surfaces \`fact_check_unverified\` status through \`/api/reviews\`, \`/api/module/{key}/state\` diagnostics, \`/api/briefing/session\` alerts, and the dashboard Review Audit panel. All of these read from the markdown audit files in \`.pipeline/reviews/*.md\`.

However, \`scripts/pipeline_v2/review_worker.py\` emits \`check_passed\`, \`check_failed\`, and \`fact_check_unverified\` events to the control plane SQLite DB — it **does not** write markdown files to \`.pipeline/reviews/\`. That path is still written by the v1 orchestrator (\`scripts/v1_pipeline.py\` → \`append_review_audit\`).

**Consequence:** v2 review runs enforce rejection correctly (per #306/#307), but their unverified claims do not appear in the API surfacing or dashboard until a v1-style audit file lands for that module. The triage signal is incomplete.

Caught by Gemini during cross-family review of #307: \"This diagnostic will primarily catch v1-originated or manually recorded unverified claims until the v2 audit writer is landed.\"

## Scope

Bridge the v2 event stream into the same markdown audit format v1 writes, so the API + dashboard surfaces become v2-complete.

Two design paths — pick one per your taste:

### Option A: teach \`review_worker\` to write markdown

On every \`check_passed\` / \`check_failed\` event, append a v1-compatible entry to \`.pipeline/reviews/{module_key}.md\` via a shared helper (factor \`append_review_audit\` out of \`v1_pipeline.py\` into a module both workers can import).

### Option B: teach \`/api/reviews\` to read from the event stream

Query \`control_plane\` events directly and synthesize the same \`fact_check_status\` derivation from event payloads instead of parsing markdown. Markdown files become optional / legacy.

## Acceptance Criteria

- [ ] A v2 review producing FACT_CHECK with \`unverified:\` evidence shows \`fact_check_status: unverified\` in \`/api/reviews\` within one poll cycle
- [ ] \`/api/module/{key}/state\` diagnostics surface \`fact_check_unverified\` for v2-rejected modules
- [ ] Dashboard Review Audit panel shows v2-originated unverified pills
- [ ] No duplicate writes if both v1 and v2 ever run on the same module (de-dup by lease/job_id)
- [ ] \`.venv/bin/pytest\` passes; \`ruff check\` clean

## References

[...]

PR #307, opened 2026-04-18
fix(pipeline): reject FACT_CHECK with unverified evidence; surface status (#306)

krisztiankoos:
## Summary
- Closes #306. FACT_CHECK results with \`evidence\` starting \`unverified:\` now route to \`patch\` and land as \`rejected\` instead of being silently approved while only emitting telemetry.
- \`/api/reviews\` gains \`fact_check_status\` + \`unverified_evidence[]\` plus a \`?fact_check_status=\` filter; \`/api/module/{key}/state\` diagnostics surface \`fact_check_unverified\`; \`/api/briefing/session\` alerts and action_rows include unverified modules.
- Dashboard gains a Review Audit panel with colored status pills and a snippet of the unverified claim.

## Test plan
- [x] \`pytest tests/test_pipeline_v2_review.py::test_review_worker_enforces_review_outcomes tests/test_local_api.py::test_reviews_index_and_single tests/test_local_api.py::test_reviews_route_filter_module_state_diag_and_briefing_unverified\` — 3 passed
- [x] \`ruff check\` — clean on changed files
- [ ] Cross-family review: needs Gemini or Claude adversarial pass (Codex drafted; rule 10 in AGENTS.md)

## Notes
- Fact-check status is derived heuristically from the markdown review log (parses \`- **FACT_CHECK**:\` lines and \`unverified:\` prefixes). Deterministic for the current \`.pipeline/reviews/*.md\` schema; will need revisiting if the review audit artifact schema changes to JSON.
- Net +113 LOC, under the 150-LOC target from the issue.

🤖 Drafted by Codex via agent bridge; reviewed and landed by Claude.

PR #309, opened 2026-04-18
fix(pipeline): bridge v2 review audit writes (#308)

krisztiankoos:
## Summary
- bridge v2 review outcomes into `.pipeline/reviews/*.md` via a small shared audit helper
- write v1-compatible `REVIEW` entries after v2 `check_passed`/`check_failed`, including job/lease ids for de-dup
- cover verified, unverified, and duplicate-write behavior in the v2 review-worker tests

## Tests
- `ruff check scripts/pipeline_common/review_audit.py scripts/pipeline_v2/review_worker.py tests/test_pipeline_v2_review.py`
- `pytest tests/test_pipeline_v2_review.py tests/test_local_api.py tests/test_v1_pipeline_retry_policy.py`

Refs: #308, #307
LOC: 118 insertions / 3 files

### DIFF
--- a/scripts/pipeline_common/review_audit.py
+++ b/scripts/pipeline_common/review_audit.py
@@ -1,0 +1,71 @@
+from __future__ import annotations
+
+import fcntl
+import re
+from datetime import UTC, datetime
+from pathlib import Path
+def _render_audit_entry(event: str, timestamp: str, fields: dict) -> str:
+    checks = fields.get("checks") if isinstance(fields.get("checks"), list) else []
+    passed = [str(check.get("id", "?")) for check in checks if isinstance(check, dict) and check.get("passed")]
+    failed = [str(check.get("id", "?")) for check in checks if isinstance(check, dict) and not check.get("passed")]
+    lines = [
+        f"## {timestamp} — `{event}` — `{fields.get('verdict', '')}`",
+        "",
+        f"**Reviewer**: {fields.get('reviewer', 'unknown')}",
+        f"**Attempt**: {fields.get('attempt', '?')}",
+        f"**Severity**: {fields.get('severity', 'unknown')}",
+        f"**Checks**: {len(passed)}/{len(checks)} passed" + (f" ({' '.join(passed)})" if passed else "") + (f" | **Failed**: {' '.join(failed)}" if failed else ""),
+        f"**Job Id**: {fields['job_id']}",
+        f"**Lease Id**: {fields['lease_id']}",
+    ]
+    failed_evidence = [
+        f"- **{check.get('id', '?')}**: {str(check.get('evidence', '')).strip()}"
+        for check in checks
+        if isinstance(check, dict) and not check.get("passed") and str(check.get("evidence", "")).strip()
+    ]
+    if failed_evidence:
+        lines.extend(["", "**Failed check evidence**:", *failed_evidence])
+    feedback = str(fields.get("feedback", "")).strip()
+    if feedback:
+        lines.extend(["", "**Feedback**:", *[f"> {line}" if line else ">" for line in feedback.splitlines() or [feedback]]])
+    return "\n".join(lines)
+def append_review_audit(module_path: Path, event: str, **fields) -> Path:
+    repo_root = next((parent for parent in (module_path.parent, *module_path.parents) if (parent / ".pipeline").exists()), Path.cwd())
+    module_key = str(fields.pop("module_key", module_path.stem)).removesuffix(".md")
+    target = repo_root / ".pipeline" / "reviews" / f"{module_key.replace('/', '__')}.md"
+    timestamp = fields.pop("timestamp", None)
+    if not isinstance(timestamp, str):
+        current = timestamp or datetime.now(UTC)
+        timestamp = (current.replace(tzinfo=UTC) if current.tzinfo is None else current.astimezone(UTC)).strftime("%Y-%m-%dT%H:%M:%SZ")
+    target.parent.mkdir(parents=True, exist_ok=True)
+    with open(target.with_suffix(".lock"), "w", encoding="utf-8") as lock_file:
+        fcntl.flock(lock_file, fcntl.LOCK_EX)
+        try:
+            existing = target.read_text(encoding="utf-8") if target.exists() else ""
+            if any(fields.get(key) and f"**{key.replace('_', ' ').title()}**: {fields[key]}" in existing for key in ("job_id", "lease_id")):
+                return target
+            body = re.split(r"\n---\n+", existing, maxsplit=1)[1].lstrip() if "\n---\n" in existing else ""
+            entry = _render_audit_entry(event, timestamp, fields)
+            display_path = str(module_path.resolve()) if repo_root not in module_path.resolve().parents else str(module_path.resolve().relative_to(repo_root))
+            target.write_text(
+                "\n".join([
+                    f"# Review Audit: {module_key}",
+                    "",
+                    f"**Path**: `{display_path}`",
+                    f"**First pass**: {(re.findall(r'^## ([0-9TZ:\\-]+) — ', existing, re.MULTILINE) or [timestamp])[-1] if existing else timestamp}",
+                    f"**Last pass**: {timestamp}",
+                    f"**Total passes**: {len(re.findall(r'^## ([0-9TZ:\\-]+) — ', existing, re.MULTILINE)) + 1}",
+                    "**Current phase**: pending",
+                    "**Current reviewer**: -",
+                    "**Current severity**: -",
+                    "",
+                    "---",
+                    "",
+                    entry if not body.strip() else f"{entry}\n\n---\n\n{body.strip()}",
+                    "",
+                ]),
+                encoding="utf-8",
+            )
+        finally:
+            fcntl.flock(lock_file, fcntl.LOCK_UN)
+    return target
--- a/scripts/pipeline_common/review_audit.py
+++ b/scripts/pipeline_common/review_audit.py
@@ -1,71 +1,147 @@
 from __future__ import annotations
 
 import fcntl
+import os
 import re
 from datetime import UTC, datetime
 from pathlib import Path
+
+
+def _atomic_write_text(path: Path, content: str) -> None:
+    """Stage to a unique tempfile in the same dir, then os.replace — survives SIGKILL mid-write."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    unique = f".{os.getpid()}.{datetime.now(UTC).strftime('%H%M%S%f')}.tmp"
+    tmp = path.with_suffix(path.suffix + unique)
+    try:
+        tmp.write_text(content, encoding="utf-8")
+        os.replace(tmp, path)
+    except Exception:
+        if tmp.exists():
+            try:
+                tmp.unlink()
+            except OSError:
+                pass
+        raise
+
+
+def _format_duration(seconds) -> str:
+    if seconds is None:
+        return "unknown"
+    try:
+        total = max(0.0, float(seconds))
+    except (TypeError, ValueError):
+        return str(seconds)
+    if total >= 60:
+        minutes = int(total // 60)
+        remainder = int(round(total - minutes * 60))
+        if remainder == 60:
+            minutes, remainder = minutes + 1, 0
+        return f"{minutes}m {remainder}s"
+    if total >= 10:
+        return f"{int(round(total))}s"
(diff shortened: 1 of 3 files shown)


krisztiankoos:
**Review** (gemini-3-flash-preview)

I will begin by researching the existing `v1_pipeline.py` implementation to determine if the new shared helper correctly replaces the original logic or introduces duplication, and I will verify the de-duplication and prepend logic in `review_audit.py` to ensure consistency with the established audit format.
I will now examine the audit writing logic in `v1_pipeline.py` to compare it with the new `review_audit.py`, and I'll also inspect `local_api.py` to understand how `_LATEST_REVIEW_RE` is used for parsing these audits.
I'll search for "Review Audit" in `scripts/v1_pipeline.py` to locate the existing audit writing logic, as my previous search for the file path was unsuccessful.
The `review_audit.py` implementation correctly prepends entries while maintaining the header, mirroring `v1_pipeline.py`'s behavior, and utilizes `fcntl.flock` for concurrency. However, I noticed that `review_audit.py` hardcodes the current state fields, unlike `v1_pipeline.py` which pulls them from `load_state()`, and since `v1_pipeline.py` hasn't been updated to use the new helper, we have immediate logic drift. I'll now check `_LATEST_REVIEW_RE` in `local_api.py` to ensure it's compatible with this format.
The `local_api.py` regex correctly matches the "REVIEW" entry format generated by `review_audit.py`. However, I need to verify if `review_worker.py` prepends the "unverified:" prefix to fact claims, as expected by `local_api.py`. I'll now examine `scripts/pipeline_v2/review_worker.py` to see how `unverified_fact_claims` is populated.
`review_worker.py` correctly handles "unverified" fact claims by appending them to the audit feedback, ensuring compatibility with `local_api.py`. However, `review_audit.py` hardcodes the "Current phase," "reviewer," and "severity" in the header, which will overwrite potentially accurate information from v1 with "pending" and "-" when v2 runs. This is especially problematic because `v1_pipeline.py` hasn't been updated to use the shared helper, leading to two different implementations writing to the same files. I'll now examine the locking logic in `v1_pipeline.py` to check for further inconsistencies.
`v1_pipeline.py` already contains an `append_review_audit` function, and the PR introduces a new, slightly different version in `scripts/pipeline_common/review_audit.py`, which is exactly the "drift problem" the PR meta warned about. The new version hardcodes header fields that v1 dynamically populates and uses a different method for determining the module key and path. Furthermore, the new de-duplication logic is unique to v2, and since v1 hasn't been updated to use the shared helper, the two versions will inconsistently manage the audit logs. I'll now examine `_render_audit_entry` in `v1_pipeline.py` to see if it includes the `Job Id` or `Lease Id` fields.
I'll examine the remaining part of `_render_audit_entry` for the `REVIEW` event in `scripts/v1_pipeline.py`.
`v1_pipeline.py`'s `_render_audit_entry` doesn't write `Job Id` or `Lease Id`, so v1 entries won't be de-duped by the new v2 logic, though v2 retries will be. While `local_api.py` correctly pulls the latest review status from the individual entries, bypassing the header, the hardcoded "pending" phase in the new v2 header still poses a risk of confusing the dashboard and introduces immediate logic drift. The duplication of `append_review_audit` means they'll likely diverge silently. I'll now examine `tests/test_pipeline_v2_review.py` to see how these changes are being verified.
The `test_review_worker_audit_dedupes_same_job` test from the PR diff seems to be missing in my current read of `tests/test_pipeline_v2_review.py`, possibly due to truncation or file versioning. I'll re-examine the lines around 318 to confirm.
I'm reviewing the provided diff for PR #309, specifically the new `append_review_audit` in `scripts/pipeline_common/review_audit.py`. The `module_key` logic correctly uses `lease.module_key` when provided, but its fallback to `module_path.stem` is flawed for nested modules. The `_render_audit_entry` closely mirrors v1, but its feedback section uses a manual blockquote implementation instead of `v1_pipeline.py`'s `_audit_quote_block`. I'll now check `v1_pipeline.py` to see how `_audit_quote_block` is implemented.
The manual blockquote implementation in `review_audit.py` matches `v1_pipeline.py`, ensuring consistent visual output. However, `review_audit.py`'s hardcoded header fields are a regression if it's meant to be a shared helper, and the fact that `v1_pipeline.py` remains unupdated makes the "factoring" incomplete and risky. I also noticed that the new helper lacks the `Duration` field present in v1, and its non-atomic `write_text` is less safe than v1's `_atomic_write_text`. I'll now examine the `_atomic_write_text` implementation in `v1_pipeline.py` to confirm.
`v1_pipeline.py` uses a superior `_atomic_write_text` that avoids concurrency races and partial writes, a feature missing from the new `review_audit.py`. The "shared helper" actually introduces duplication, and its lack of error handling in `review_worker.py` could block pipeline progress if I/O fails. Furthermore, the timestamp logic in `review_audit.py` is unnecessarily complex compared to v1's helper. I'll now check `v1_pipeline.py`'s `_format_audit_timestamp` to see how it compares.
The timestamp logic in `review_audit.py` is essentially the same as in `v1_pipeline.py`, just inlined. I've noted that `review_worker.py` now includes `review_model` in the review results and appends to the audit. However, `review_worker.py` skips the audit append if preflight fails, whereas v1 typically logs such failures. I'll verify this discrepancy by checking `v1_pipeline.py`'s behavior for deterministic check failures.
I'll search for calls to `append_review_audit` in `scripts/v1_pipeline.py` to understand when v1 logs audit entries.
`append_review_audit` is called within `emit_audit`, which itself is a helper in `_run_module_cycle`. I'll now find where `emit_audit` is called to trace the audit logging flow.
The new `review_audit.py` helper is a step backwards: it lacks the event-specific formatting, atomic write safety, and detailed metadata of v1, and its hardcoded "pending" status will actively corrupt v1's audit logs. Furthermore, the failure to update `v1_pipeline.py` ensures immediate logic drift, and the omission of preflight logging in v2 creates a visibility gap compared to v1. I'll now check `local_api.py` for `build_module_reviews` to see how the test verification works.
The test confirms `local_api.py` successfully parses v2-written audits. I'll now check the start of `build_module_reviews` in `local_api.py` to see if it relies on the audit header.
`build_module_reviews` doesn't seem to parse the header. I'll find where `fact_check_status` is derived to confirm.
`build_module_reviews` uses `_fact_check_summary` and `_LATEST_REVIEW_RE` to find the newest `REVIEW` entry, which works because new entries are prepended. I noticed a potential issue with `re.split` in `review_audit.py` if the content contains `---`, but since `v1_pipeline.py` uses similar logic, I'll check its `_split_audit_header_body` to confirm how it handles separators.
`v1_pipeline.py` splits at the first `\n---\n`, which works correctly. While the PR's `review_audit.py` works for v2, it's a flawed "lazy" unification that copies and degrades logic. Most critically, hardcoding "pending" in the header will corrupt v1's status tracking. I'll issue a **NEEDS CHANGES** verdict, requiring the helper to preserve header metadata, support the `Duration` field, use atomic writes, and include preflight logging for consistency. Updating `v1_pipeline.py` to use the helper would also prevent immediate drift and likely stay within the LOC budget.

1. Verdict: NEEDS CHANGES

2. Specific feedback:

- **scripts/pipeline_common/review_audit.py:53-55**: **CRITICAL REGRESSION**. Hardcoding "Current phase: pending", "Current reviewer: -", and "Current severity: -" will overwrite accurate v1 metadata whenever a v2 review runs. The helper MUST preserve these values from the existing `header` if present, or accept them as arguments.
- **scripts/pipeline_common/review_audit.py:45-66**: **REDUCED SAFETY**. The implementation uses `target.write_text()` inside a lock but lacks the atomic "write-to-tmp-then-rename" pattern found in `v1_pipeline.py`'s `_atomic_write_text`. A crash or disk failure during write will leave the audit file truncated. 
- **scripts/pipeline_common/review_audit.py:20**: **MISSING FIELD**. The `_render_audit_entry` helper omits the `Duration` field present in v1 audits. This prevents the dashboard from surfacing performance metrics for v2 runs.
- **scripts/pipeline_v2/review_worker.py:65-75**: **VISIBILITY GAP**. Preflight failures are emitted to the control plane but are NOT appended to the markdown audit. v1 logs `CHECK_FAIL` events; v2 should do the same to ensure the audit log is a complete record of the module's quality history.
- **scripts/pipeline_common/review_audit.py:33**: **NESTED MODULE BUG**. The fallback `module_path.stem` loses directory information (e.g., `docs/k8s/intro.md` becomes `intro`). While `review_worker.py` passes the key correctly, the helper's default is dangerous for other callers.

3. Reality Check:
The PR claims to "bridge" v2 audits but introduces a parallel, incompatible implementation of `append_review_audit` that will cause the two versions to fight over the file header. The LOC budget excuse for not updating `v1_pipeline.py` is invalid; removing the duplicate logic from `v1_pipeline.py` and importing the helper would improve the project's health without significantly increasing the diff size.

Recommendation: **BLOCK** until header preservation and atomic writes are implemented. Moving `_atomic_write_text` to `pipeline_common` and having both v1 and v2 use it is the only way to prevent silent corruption.

krisztiankoos:
**Review** (gemini-3-flash-preview)

Verdict: NEEDS CHANGES

The commit addresses the structural issues (atomic writes, header preservation, duration schema), but the implementation in `review_worker.py` is incomplete and contains a breaking regression.

1. **CRITICAL (Regression)**: `scripts/pipeline_v2/review_worker.py` is missing the import for `append_review_audit`. The worker will crash on preflight failure.
2. **MISSING (Visibility)**: The fix only logs `CHECK_FAIL` (preflight). Successful or failed LLM reviews (the primary "REVIEW" event) are still not logged to the markdown audit, leaving the visibility gap largely open.
3. **SAFETY (Fixed)**: `_atomic_write_text` implementation is correct and robust.
4. **INTEGRITY (Fixed)**: `_parse_existing_header` successfully preserves v1 metadata, allowing coexistence without data loss.
5. **MINOR (Fixed)**: Removing the `module_path.stem` fallback correctly forces explicit module keys.

Regarding the deferred factoring: **APPROVE as follow-up.** The header preservation logic in `review_audit.py` effectively mitigates the data corruption risk, so merging v1/v2 logic into a single shared function is no longer a safety blocker for this PR.

**Actionable fix**: Add the import to `review_worker.py` and add an `append_review_audit` call after `self._run_llm_review` to capture deep review outcomes.

krisztiankoos:
## Rebuttal on iteration-1 review

Two of the five items flagged as regressions are false positives — the code already does what the review says is missing. Evidence:

### "CRITICAL: missing import for append_review_audit"

\`scripts/pipeline_v2/review_worker.py\` line 10:

\`\`\`
from pipeline_common.review_audit import append_review_audit
\`\`\`

Import is present. The worker does not crash on preflight — covered by \`test_review_worker_enforces_review_outcomes\` (which hits the non-preflight path) and will be extended if needed.

### "MISSING: REVIEW events not logged to markdown"

Codex's original commit \`43184b6d\` (the base of this PR) already wired \`append_review_audit\` for the REVIEW outcome at \`review_worker.py:172-184\`. Iteration 1 added a separate call for the preflight branch — it did not remove the existing REVIEW logging. You can see both calls in the current file.

### Test evidence

\`\`\`
PYTHONPATH=scripts pytest tests/test_pipeline_v2_review.py
============================== 10 passed in 0.24s ==============================
\`\`\`

Including the new \`test_audit_preserves_v1_written_header_metadata\` which specifically exercises both v1 header preservation and v2 REVIEW append.

### Where we stand on the five items

| # | Item | Status |
|---|------|--------|
| 1 | Header preservation | Fixed (confirmed in follow-up) |
| 2 | Atomic writes | Fixed (confirmed) |
| 3 | Duration field | Fixed (confirmed) |
| 4 | Preflight logging | Added (correctly flagged as new behaviour in this iteration) |
| 5 | Module key fallback | Fixed (confirmed) |

All five substantive items addressed. Re-dispatching for a fresh read.

krisztiankoos:
**Review** (gemini-3-flash-preview)

1. Clear verdict: APPROVE

2. Specific, actionable feedback:
- CLAIM 1 status: **TRUE** (scripts/pipeline_v2/review_worker.py:10).
- CLAIM 2 status: **TRUE** (scripts/pipeline_v2/review_worker.py:185).
- The implementation correctly distinguishes between deterministic failures (`"CHECK_FAIL"`, line 96) and LLM review results (`"REVIEW"`, line 187). 
- Note that line 185 properly captures the main review outcome after LLM processing, including enriched feedback for unverified claims (lines 181-184).

3. Concise — The requested audit logging functionality is present and correctly distinguishes between pipeline phases. No issues found.

Issue #279, opened 2026-04-18
Automated citation pipeline: Gemini 3.1 Pro writes → Codex fact-checks and reviews


krisztiankoos:
## Context

Policy: `docs/citation-upgrade-plan.md` (committed b249f710). Deterministic checker: `scripts/check_citations.py` (49161266). Status endpoint: `/api/citations/status` (9cc52d9d). Seed data for first batch: `docs/citation-seeds-ai-foundations.md` (5cfa70b0).

What's missing: the actual pipeline wiring that turns seeds → module with citations → reviewed → merged.

## Scope

Extend `scripts/v1_pipeline.py` with a citation-aware WRITE → REVIEW path:

1. **WRITE (Gemini 3.1 Pro)**: inject citation seeds for the target module (lookup by `module_key` against `docs/citation-seeds-*.md`). Writer prompt already carries fact-ledger claims; add a new `## Authoritative Sources` block with the pre-vetted URLs the writer MUST cite inline.
2. **CITATION CHECK**: run `scripts/check_citations.py` as a hard gate. If `< N` citations or dead URLs → reject, loop back to WRITE with the failure reason.
3. **REVIEW (Codex)**: existing binary-gate review, PLUS a new `citations` dimension scored against `docs/quality-rubric.md`. A module cannot score `clean` with citations < 3/5.
4. **MERGE**: unchanged.

First batch target: AI modules 1.1, 1.2, 1.3 (seeds already staged). This is the rollback that triggered the whole effort — it must land citation-first this time.

## Acceptance Criteria

- [ ] `scripts/v1_pipeline.py` reads citation seeds from `docs/citation-seeds-{track}.md` and injects into WRITE prompt
- [ ] `step_check_citations()` added between WRITE and REVIEW, calls `scripts/check_citations.py`, loops back to WRITE on failure
- [ ] `citations` dimension added to the review rubric; REVIEW prompt updated
- [ ] Pipeline e2e run on AI modules 1.1, 1.2, 1.3 produces drafts scoring `clean` with citations ≥ 3/5
- [ ] `docs/citation-seeds-ai-foundations.md` fully consumed (all 13 citations surface in the final drafts, measured by `check_citations.py`)
- [ ] `python -m unittest scripts.test_pipeline` passes (new tests cover citation gate)
- [ ] `/api/citations/status` reflects the new drafts without code change
[...]

PR #310, opened 2026-04-18
feat(pipeline): inject write-time citation seeds (#279)

krisztiankoos:
Subset of #279, part 1 of 3: implements #279a.

## Summary
- derive `citation-seeds-{track}.md` from the current `module_key` during WRITE
- extract only the matching module section and inject it into the WRITE prompt as `## Authoritative Sources — cite these inline`
- leave WRITE unchanged when the seeds file is missing or the module has no seeded entry
- add unit coverage for match, missing file, and missing module cases

## Verification
- `../../.venv/bin/python -m unittest scripts.test_pipeline` (passes with the 2 known pre-existing failures allowed by AGENTS: `check_failures` tests)
- `npm run build` from primary checkout: passes
- `ruff check scripts/v1_pipeline.py scripts/test_pipeline.py` could not be validated with the required repo venv because `.venv/bin/ruff` is absent in this environment; the available global `ruff` reports existing repo-level violations in these files unrelated to this PR

Refs: #279, #279a

### DIFF
--- a/scripts/test_pipeline.py
+++ b/scripts/test_pipeline.py
@@ -2579,6 +2579,11 @@ class TestKnowledgeCards(unittest.TestCase):
         - Say VPA recommendations come from Prometheus instead of the VPA recommender path.
         """)
 
+    def _write_seed_file(self, content: str) -> None:
+        docs_dir = self.repo_root / "docs"
+        docs_dir.mkdir(parents=True, exist_ok=True)
+        (docs_dir / "citation-seeds-ai-foundations.md").write_text(content)
+
     def test_knowledge_card_written_with_correct_schema(self):
         """Generator writes a schema-valid knowledge card to the cache."""
         import generate_knowledge_card as g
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Gemini cross-family review (dispatch crashed mid-post — relaying manually)

Verdict: **NEEDS CHANGES**

### Must-fix
- **scripts/v1_pipeline.py:1216-1221** — `authoritative_sources_block` is computed in `step_write` but is **not** passed to the `REWRITE_PROMPT_TEMPLATE.format()` call in the `if rewrite:` branch. Full rewrites — the most important writing task — won't receive the citation seeds. The entire WRITE-time injection is effectively a no-op for rewrites.
- **scripts/v1_pipeline.py:~1042** — `REWRITE_PROMPT_TEMPLATE` has no `{authoritative_sources_block}` placeholder. Add it to match `WRITE_PROMPT_TEMPLATE`.
- **scripts/v1_pipeline.py:~978** — `track` derivation: `"-".join(module_key.split("/")[:-1])` yields invalid filename `docs/citation-seeds-.md` when `module_key` has no slashes (root-level modules). Either skip injection gracefully or derive a sensible default (e.g. `citation-seeds-root.md`).

### Nit
- **scripts/v1_pipeline.py:~1007** — URL extraction is `line.startswith("- ") and "http" in line`. Works for the current bullet-style seeds file, but will miss numbered lists or raw URLs if the seed-file format ever broadens. Either document the constraint in the seed-file schema or widen the parser.

### Scope discipline
Correctly stayed inside #279a — no step_check_citations, no rubric changes. **However**, the rewrite-injection omission means #279a doesn't actually deliver WRITE-time seed injection end-to-end as the issue spec required.

(Review posted manually — dispatch script crashed after producing this output because its cwd was in a just-deleted worktree.)

krisztiankoos:
**Review** (gemini-3-flash-preview)

1. Clear verdict: APPROVE

2. Specific, actionable feedback:
All four must-fix items from the initial review have been correctly resolved. Specifically:
- **Fix 1 & 2**: `authoritative_sources_block` is now defined in `step_write`, passed to the `.format()` call, and the corresponding placeholder has been added to `REWRITE_PROMPT_TEMPLATE`.
- **Fix 3**: `_citation_seed_path` now correctly handles root-level module keys (e.g., `module-0.1-test.md`) by returning `None` instead of generating a broken path like `citation-seeds-.md`.
- **Fix 4**: URL extraction logic in `_format_authoritative_sources_for_prompt` has been expanded from simple bullets to also support numbered lists (`1. http...`) and raw URLs (`http...`) using `re.match` and `line.startswith`.

3. Concise — focus on what needs changing:
The implementation is solid and verified by the new test cases in `scripts/test_pipeline.py`. No further changes required.

Issue #322, opened 2026-04-20
Pipeline v4: content-first (expand to 4/5) then citation


krisztiankoos:
## Context

Session 7 (2026-04-19 → 2026-04-20) landed `pipeline_v3` and ran it over
ZTT (6/6) and the /ai track (25/25). All green at the citation/gate level —
but end-of-session surfaced a structural problem: /ai modules are thin
(129–203 LOC vs rubric's 600–800+ minimum) and most are missing the
required Quiz section. Rubric scores:

| Subsection | Avg | Range |
|---|---:|---|
| ai-building | 3.5 | [3.5–3.5] |
| ai-for-k8s | 2.0 | [2.0–2.0] |
| ai-native-work | 2.0 | [2.0–2.0] |
| foundations | 2.7 | [2.0–4.2] |
| open-models | 3.5 | [3.5–3.5] |

Primary issue on 23/25: "thin, no quiz". 2 more: "no quiz".

**Citation-first was the wrong order.** When these modules are later
expanded for rubric compliance, anchor_text references will break,
citations will anchor into paragraphs that get rewritten, and the
citation pass will largely need to be redone. The right sequence:

1. Expand to 4/5 (length, quiz, mistakes, exercises)
2. Rubric gate confirms 4/5
3. Citation pass on content-stable modules

The original design proposed "additive-only EOF append" expansion. That
was adversary-reviewed by Gemini 3.1-pro
([comment](https://github.com/kube-dojo/kube-dojo.github.io/issues/322#issuecomment-4287421505))
and came back **NEEDS CHANGES (2/5)** with five specific findings. The
revised design below addresses them.

## pipeline_v4 revised design

```
Stage 1  RUBRIC_SCAN     →  scripts/rubric_gaps.py
                            Wraps /api/quality/scores, emits structured
                            gap list per module:
                              { module, path, score, gaps: [thin|no_quiz|
                                no_mistakes|no_exercise|no_diagram|...],
                                target_loc: 600..800 }
                            Score ≥ 4.0 skips straight to Stage 4.

Stage 2  EXPAND           →  scripts/expand_module.py
                            Section-aware injection via hand-rolled
                            section splitter (see below). Per gap type:
                              gap=quiz       → Codex, 6-8 scenario Qs,
                                               injected at Quiz section
                              gap=mistakes   → Codex, 6-8-row table,
                                               injected at Common Mistakes
                              gap=exercise   → Codex, multi-step hands-on,
[...]

PR #329, opened 2026-04-21
fix(pipeline-v4): expose public skip-citation mode (#322)

krisztiankoos:
## Summary
- expose `skip_citation` on the public `run_pipeline_v4()` entrypoint
- switch `pipeline_v4_batch.py` back to the public API instead of the private `_run_pipeline_v4()` helper
- add regression coverage for score-first runs that intentionally stop after stages 1-3

## Validation
- `.venv/bin/ruff check scripts/pipeline_v4.py scripts/pipeline_v4_batch.py tests/test_pipeline_v4.py`
- `.venv/bin/python -m pytest tests/test_pipeline_v4.py tests/test_pipeline_v4_batch.py`
- `/Users/krisztiankoos/projects/kubedojo/.venv/bin/python scripts/test_pipeline.py`

## Notes
- This does not change the bounded-worker design in `pipeline_v4_batch.py`; the 8-way concurrency text in #322 is stale versus the shipped rate-limit guardrails.
- It aligns the public API with the already-shipped score-first `--skip-citation` batch path so callers no longer need a private helper.

### DIFF
--- a/scripts/pipeline_v4.py
+++ b/scripts/pipeline_v4.py
@@ -545,13 +545,14 @@ def run_pipeline_v4(
     max_rubric_retries: int = 2,
     generated_loc_threshold: float = 0.5,
     dry_run: bool = False,
+    skip_citation: bool = False,
 ) -> PipelineV4Result:
     return _run_pipeline_v4(
         module_key,
         max_rubric_retries=max_rubric_retries,
         generated_loc_threshold=generated_loc_threshold,
         dry_run=dry_run,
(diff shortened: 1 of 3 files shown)


krisztiankoos:
Gemini review:

APPROVE

No blocking findings.

**Residual risk note:** Very low risk. This PR cleanly exposes the existing `skip_citation` logic through the public API and correctly updates the batch caller, covered by an explicit regression test to ensure the citation pipeline is bypassed when the flag is set.

Issue #243, opened 2026-04-15
Translation v2: queue-based Ukrainian sync control plane


krisztiankoos:
Parent: #186
Related: #239, #241

## Problem
`uk_sync.py` is still a script-driven best-effort loop. It does not have the same operational guarantees as pipeline v2:
- no per-module job queue
- no leases / recovery surface
- no dead-letter handling
- weak visibility into in-flight vs stalled work
- section-wide runs can hang on one slow module
- translation freshness is outside the main control plane

Recent ZTT work exposed the gap directly: a chunked translation timed out, the wrapper process stalled around an orphan child, and the section run kept ownership of work without a proper queue/lease model.

## Goal
Build a translation control plane with the same operational shape as v2, but specialized for Ukrainian sync.

## Required capabilities
- queue translation jobs per module, not per whole section
- explicit phases for detect -> translate/fix -> verify -> complete
- lease-based worker execution with watchdog recovery
- strict freshness mode: commit drift counts as stale
- retries / dead-letter state for flaky or oversized modules
- status surfaces for repo-wide and per-track translation health
- small-batch execution so one slow module does not block a whole section
- preserve translation-specific checks: glossary, anti-Russicism, lab metadata parity

## Initial rollout
1. Reuse existing v2 control-plane primitives where practical.
2. Start with ZTT Ukrainian only.
3. Run in very small batches first.
4. Once stable, widen to more tracks.

## Acceptance
- ZTT Ukrainian can be processed module-by-module under queue control
- stalled jobs are visible and recoverable
- status command distinguishes synced / stale / missing / leased / dead-letter
- no section-wide opaque hangs
- repo state survives context switches cleanly

krisztiankoos:
Implemented initial translation-v2 slice locally: dedicated translation queue on , per-module enqueue/status/worker/watchdog commands, strict freshness reporting, and integration into . Focused test slice is green: . Live ZTT freshness now shows only one remaining stale Ukrainian module: .

krisztiankoos:
Implemented initial translation-v2 slice locally:
- dedicated translation queue on `.pipeline/translation_v2.db`
- per-module enqueue/status/worker/watchdog commands in `scripts/translation_v2.py`
- strict freshness reporting for missing/stale/synced state
- integration into `scripts/status.py`
- focused test slice green: `12 passed`

Live ZTT freshness now shows only one remaining stale Ukrainian module: `module-0.6-git-basics`.

[...]

PR #330, opened 2026-04-21
fix(translation-v2): expose leased jobs in status (#243)

krisztiankoos:
## Summary
- return explicit `in_progress` module lists from `translation_v2` queue status, not just counts
- add regression coverage for leased translation jobs and per-track translation-v2 visibility
- close the remaining status-surface gap in the translation-v2 control plane

## Validation
- `.venv/bin/ruff check scripts/translation_v2.py tests/test_translation_v2.py`
- `.venv/bin/python -m pytest tests/test_translation_v2.py tests/test_status_script.py tests/test_local_api.py`
- `/Users/krisztiankoos/projects/kubedojo/.venv/bin/python scripts/test_pipeline.py`

## Notes
- `#243`'s core queue/lease implementation was already on `main`; this PR finishes the remaining leased-job visibility gap so stalled work is observable in the same status surface used for recovery and rollups.
- No Astro/content changes, so `npm run build` was not required.

### DIFF
--- a/scripts/translation_v2.py
+++ b/scripts/translation_v2.py
@@ -246,6 +246,7 @@ def _build_translation_queue_status(db_path: Path) -> dict[str, Any]:
 
     pending_review_list = sorted([m for m, status in status_by_module.items() if status == "pending_review"])
     pending_write_list = sorted([m for m, status in status_by_module.items() if status == "pending_write"])
+    in_progress_list = sorted([m for m, status in status_by_module.items() if status == "in_progress"])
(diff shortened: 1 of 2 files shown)


krisztiankoos:
Gemini review:

No blocking findings.

**Residual risk:** Visibility into `in_progress` modules depends entirely on the `leased` state in the SQLite DB; if a translation process crashes without a cleanup handler, these modules will appear active indefinitely until manual lease reclamation occurs.

PR #349, opened 2026-04-23
fix(dispatch): auto-fallback from API-key to OAuth on Gemini 429

krisztiankoos:
## Summary

- When the Gemini API-key tier hits a 429 / quota error mid-dispatch, the next retry now automatically strips the key and uses the OAuth/subscription path — independent quota, so retry is immediate (no backoff).
- `KUBEDOJO_GEMINI_SUBSCRIPTION=1` is preserved as an opt-out: skip the API-key attempt entirely when that tier is known-exhausted.
- Only after the subscription path *also* rate-limits do we apply exponential backoff.

## Why

Previously `KUBEDOJO_GEMINI_SUBSCRIPTION=1` was a static, up-front switch. On a long batch, if the API-key tier hit 429 mid-run, there was no runtime fallback — the pipeline just hammered the same exhausted tier until `max_retries` ran out.

## Test plan

- [x] New unit tests in `tests/test_dispatch_gemini_timeout.py`:
  - `_gemini_env` strips keys only on subscription, envs are independent dicts
  - 429 on API-key → next retry uses `use_subscription=True`, no sleep between attempts
  - Force-subscription env var → first call already on subscription
- [x] Live dispatch smoke (`dispatch_gemini_with_retry`) returns `OK` via API-key path, no zombies
- [x] 11/11 existing dispatch tests pass; ruff clean

### DIFF
--- a/scripts/dispatch.py
+++ b/scripts/dispatch.py
@@ -40,12 +40,27 @@ _ENV["GEMINI_SESSION"] = "1"        # Disable hostile aliases (eza, bat, zoxide)
 _ENV["KUBEDOJO_PIPELINE"] = "1"     # Suppress inbox hooks during pipeline runs
 # Gemini auth: CLI prefers GEMINI_API_KEY when set; otherwise falls
 # through to OAuth/subscription creds in ~/.gemini/oauth_creds.json.
-# When the API-key tier is rate-limited, set
-# KUBEDOJO_GEMINI_SUBSCRIPTION=1 to strip the key from child env so
-# dispatches use the subscription path instead.
-if os.environ.get("KUBEDOJO_GEMINI_SUBSCRIPTION") == "1":
-    _ENV.pop("GEMINI_API_KEY", None)
-    _ENV.pop("GOOGLE_API_KEY", None)
+#
+# Runtime behavior (see dispatch_gemini_with_retry):
+#   - Default: start on API-key path. If a 429 / quota hit is detected,
+#     automatically strip the key from the child env and retry on OAuth.
+#   - Force subscription: set KUBEDOJO_GEMINI_SUBSCRIPTION=1 to skip the
+#     API-key attempt entirely (e.g. on known-exhausted API quota).
+#
+# _ENV stays unmodified (API-key path). Subscription env is built on
+# demand via _gemini_env(use_subscription=True).
+_FORCE_GEMINI_SUBSCRIPTION = os.environ.get("KUBEDOJO_GEMINI_SUBSCRIPTION") == "1"
+
+
+def _gemini_env(use_subscription: bool) -> dict:
+    """Return the child env for a Gemini dispatch. When use_subscription=True,
+    strip the API keys so the CLI falls back to OAuth creds."""
+    if not use_subscription:
+        return _ENV
+    env = {**_ENV}
+    env.pop("GEMINI_API_KEY", None)
+    env.pop("GOOGLE_API_KEY", None)
+    return env
 
 # GitHub comment char limit (65,536 minus safety margin)
 GH_CHAR_LIMIT = 64000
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Codex independent review

**Model:** `codex` (via ai_agent_bridge ask-codex --review)
**Ran:** 2026-04-23T09:45Z

Verdict: NEEDS CHANGES

FINDING: `max_retries=1` never reaches the subscription fallback after an API-key 429
FILE:LINE: scripts/dispatch.py:317
CURRENT CODE:
  for attempt in range(max_retries):
      ok, output = dispatch_gemini(prompt, model, review, timeout, mcp,
                                   use_subscription=use_subscription)
      if ok:
          return True, output

      # Check for rate limit
      if _is_rate_limited(output):
          # If we're still on the API-key path, flip to subscription and retry
          # immediately — the OAuth tier has a separate quota.
          if not use_subscription:
              print("Gemini API-key rate-limited — switching to subscription (OAuth) path",
                    file=sys.stderr)
              use_subscription = True
              continue  # no sleep; retry on the other tier
WHY: The API-key 429 still consumes a loop iteration. With `max_retries=1`, the `continue` exits the only iteration, then the function falls through to `return False, output` without ever attempting the OAuth/subscription path. That breaks the advertised "flip and retry immediately" behavior for callers using a single retry budget.
FIX: Restructure the loop so the API-key -> subscription flip does not consume the retry budget, or guarantee one subscription attempt after the first API-key 429 even when `max_retries` is 1. Add a regression test for that edge case.

Nits: none.

Concise focus: I did not find a new `_ENV` aliasing bug in the subscription path: `_gemini_env(True)` builds a fresh dict and strips keys there. I also did not find a concrete Claude/Codex regression from leaving Gemini keys in `_ENV`; those dispatchers still just inherit `_ENV` and do not consume Gemini auth vars.

---
🤖 posted by Claude Code

krisztiankoos:
## Gemini independent review

**Model:** `gemini-3-flash-preview` (via dispatch.py gemini --review)
**Ran:** 2026-04-23T10:25ZI will start by researching the `scripts/dispatch.py` file to understand the current implementation of `_is_rate_limited` and how other dispatchers like `dispatch_claude` use `_ENV`.
I will run `git diff main scripts/dispatch.py` to see the actual changes on the branch and compare them with the diff provided in the prompt.
I will read the test file `tests/test_dispatch_gemini_timeout.py` to check if the new tests correctly verify the dynamic auth fallback logic.
I will grep for `_ENV` in `scripts/dispatch.py` to ensure it is not being mutated elsewhere, which could lead to unexpected behavior in the dynamic auth fallback logic.
Verdict: APPROVE

The changes successfully implement dynamic authentication fallback for Gemini 429 errors. By moving from a static import-time environment modification to a runtime check, the system now automatically attempts to use the subscription (OAuth) path when the API-key tier is exhausted, without requiring a process restart.

The implementation of the nested call in `dispatch_gemini_with_retry` is particularly robust as it ensures that even with `max_retries=1`, both authentication tiers are attempted before failure. The test coverage is excellent, explicitly verifying the auth-flip logic, the forced subscription path, and the edge case of dual-tier rate limiting.

### Must-Fix
None.

### Nits
FINDING: Dead code outside the retry loop
FILE:LINE: `scripts/dispatch.py`:351
CURRENT CODE:
```python
    return False, output
```
WHY: This line is unreachable because the `for` loop contains a catch-all `return False, output` at its tail (line 347) for any cases that aren't continued via rate-limit backoff.
FIX: Remove the redundant return statement or the one at the end of the loop.

FINDING: Inconsistent return type in `_gemini_env`
FILE:LINE: `scripts/dispatch.py`:59
CURRENT CODE:
```python
    if not use_subscription:
        return _ENV
```
WHY: Returning the global `_ENV` by reference is inconsistent with the `True` branch which returns a copy (`{**_ENV}`). While currently safe as `subprocess.Popen` does not mutate the environment, always returning a copy provides better isolation against accidental future mutations.
FIX: Change to `return {**_ENV}`.

FINDING: Risk of false positives with "capacity" pattern
FILE:LINE: `scripts/dispatch.py`:100
CURRENT CODE:
```python
    "capacity",
```
WHY: The substring "capacity" is quite generic and could appear in legitimate technical output (e.g., "cluster capacity", "volume capacity"). 
FIX: Consider using a more specific pattern like `"No capacity available"` (which is already present) or `"model is overloaded.*capacity"`. However, since this only triggers a fallback to an independent quota, the impact of a false positive is low.

---
🤖 posted by Claude Code

Issue #343, opened 2026-04-22
citation-residuals: core auto-resolver for needs_citation findings


krisztiankoos:
Parent epic: #341

## Problem

The 276 findings (growing as more batches land) in `.pipeline/v3/human-review/*.json` under the `queued_findings.needs_citation[]` path are structured data: each has `excerpt`, `signals`, `search_hint`. Most of them are resolvable by an LLM + URL validator loop. A human doing this 1-by-1 would take hours per module. An auto-resolver should clear ≥80% of them.

## Shape of one finding

```json
{
  "line": 510,
  "signals": ["year_reference", "price_usd", "attribution"],
  "excerpt": "**Did You Know?** In November 2023, New Relic was taken private by Francisco Partners and TPG in a massive transaction valued at \$6.5 billion. According to New Relic's self-published 2026 data report, their AIOps users ship code at an 80% higher frequency.",
  "search_hint": [
    "New Relic November 2023 taken private Francisco Partners TPG 6.5 billion",
    "New Relic 2026 data report AIOps users ship code 80% higher frequency",
    "New Relic acquisition press release 2023 6.5 billion"
  ]
}
```

## Deliverable

New `scripts/citation_residuals.py` with CLI:

```bash
# Resolve residuals for a single module
python scripts/citation_residuals.py resolve <module-key>

# Bulk over everything in .pipeline/v3/human-review/
python scripts/citation_residuals.py resolve --all

# Dry-run (proposes URLs + markers but writes nothing)
python scripts/citation_residuals.py resolve <module-key> --dry-run

# Report-only (summary of queue, no resolution)
python scripts/citation_residuals.py report
```

## Inner loop per finding

1. **Candidate search** — dispatch_gemini with a structured prompt using `excerpt + search_hint` → asks for 1–3 candidate URLs with tier preference (official vendor / standards body / peer-reviewed press over blogs).
2. **Validate** — `fetch_citation.fetch(url)` on each candidate; reject tier-4 / dead / off-topic.
3. **Match content** — confirm the fetched page contains the factual anchor from `signals` (e.g. if `price_usd: \$6.5 billion`, the page should mention that figure or similar). If content doesn't match the claim, reject.
4. **Inject** — append an inline marker like `[cite:<short-id>]` at end of the flagged line, and add the URL to the module's `## Sources` section with the matching short-id.
[...]

PR #357, opened 2026-04-23
feat(#356 part 1): HEAD-check + allowlist pre-filter for LLM-suggested URLs

### LINKED ISSUE #356: #343 phase-1.5: HEAD pre-check + quote-match verification + Codex fallback (URL hallucination mitigation)
Child of #343. Addresses the main blocker surfaced by the 2026-04-23 3-module pilot: **Gemini hallucinates candidates in two ways** — (1) URLs that 404, and (2) URLs that 200 but don't actually support the claim. Both together cap phase-1's auto-resolve rate at ~20% instead of the target ≥60%.

## Pilot evidence (commit `55e0e9a2`)

Three modules, 5 needs_citation findings, current resolver:

| Module | Resolved | Unresolvable | Notes |
|---|---:|---:|---|
| advanced-genai/module-1.1-fine-tuning-llms | 1 | 1 | Wiki Algorithmic_bias worked (real URL). Gemini hallucinated `Gender_bias_in_machine_learning` — 404. |
| ai-infrastructure/module-1.1-cloud-ai-services | 0 | 2 | One `no_candidates_returned`; one `all_candidates_failed` (hallucinated Azure deprecation URL, fake Oracle press date). |
| cloud-advanced-operations/module-8.1-multi-account | 0 | 1 | Anchorless — correctly rejected (working as designed). |

Two distinct hallucination classes observed:

- **URL-existence hallucination** — Gemini invents plausible slugs. Examples: `en.wikipedia.org/wiki/Gender_bias_in_machine_learning` (404), `oracle.com/news/…-2026-03-31/` (future-dated, fake).
- **URL-content hallucination** — URL exists (200, on allowlist), but the page doesn't actually mention the claim. Today's resolver mitigates this only for findings with numeric anchors (year/price/percent); for other findings, a 200 + on-list page is accepted without verifying relevance. The current `arxiv.org/abs/2004.12338` rejection in module 1.1 cloud-ai-services was a lucky hit — the paper existed but didn't contain the claim's year.

Neither can be fixed by asking Gemini to self-verify (the failure mode is confident fabrication — self-verify extends the hallucination to the check itself). Both have to be caught deterministically outside the model.

### Part 1 — HEAD-check pre-screen (~40 LOC, 0 LLM spend)

In `request_candidates()`, after parsing Gemini's JSON response, filter candidates by `urllib.request.Request(url, method='HEAD')` with a 5s timeout. Keep only 2xx / 3xx. Eliminates URL-existence hallucinations before any allowlist / fetch / anchor-match path runs.

Fallback: some well-behaved sites reject HEAD (405) but accept GET. Retry with `GET` + `Range: bytes=0-0` on 405. Pattern already used in `fetch_citation.py`.

### Part 2 — Quote-match verification (~50 LOC, minor prompt change)

Currently the LLM prompt returns `{url, tier, why}`. Extend to return `{url, tier, why, expected_quote}` — a specific sentence-or-phrase from the page that supports the claim. The resolver then:

1. Fetches the page (current `fetch_citation.fetch` path).
2. After passing allowlist + HEAD, substring-matches `expected_quote.lower()` against the cached text body.
3. Reject with `reason=\"quote_not_in_page\"` if not found.

This catches URL-content hallucinations for findings of ALL signal types — including `named_incident` / `attribution` / `vendor_since` where our regex anchor extraction yields nothing. A claim about 'Amazon's 2014 recruiting tool' currently resolvable only via a year anchor becomes resolvable with e.g. `expected_quote=\"Amazon scrapped the project after\"` — Gemini commits to a specific sentence it believes exists, and we verify.

Quote-match is lenient: we look for a ≥12-char substring of the quote (case-insensitive, whitespace-normalized), not an exact-string match. Handles minor LLM transcription drift while rejecting fabricated quotes.

### Part 3 — Codex fallback (~80 LOC, ~10-20% extra LLM spend)

When Gemini's filtered candidate list is empty after Parts 1+2, re-request from Codex. Different training distribution → different URL knowledge → some findings Gemini can't cite, Codex can. Per review-protocol memory and PR #350 learning, Codex has caught things Gemini missed before.

Implementation: `request_candidates_with_fallback(finding)` that tries `dispatch_gemini` first, and only invokes `dispatch_codex` when Gemini returns empty or all-filtered. Apply Parts 1 + 2 to Codex candidates too.

## Expected effect on resolve rate

Theoretical ceiling with all three:
- URL-existence hallucinations → eliminated by Part 1
- URL-content hallucinations → caught by Part 2, plus anchorless findings become resolvable (Part 2 substitutes for regex anchors for non-numeric signals)
- Residual findings where both LLMs can't find a real citation → Part 3 adds ~10-20% recovery via Codex fallback

Best estimate: **60-75% auto-resolve**, up from current 20%, consistent with #343's phase-1 target.

## Acceptance

- [ ] Part 1: HEAD-check wrapper added, 405→GET-range-0 retry path. Unit tests: happy 200, 404-filtered, 405→GET success, timeout filtered, redirect-to-off-allowlist filtered.
- [ ] Part 2: prompt updated to require `expected_quote`, resolver validates quote substring-present in cached body (case-insensitive, whitespace-normalized, ≥12-char match). Unit tests: quote present (pass), quote absent (`quote_not_in_page` reason), LLM omits quote field (reject candidate).
- [ ] Part 3: Codex fallback invoked only when Gemini's filtered list is empty — don't double-spend on every finding. Apply Parts 1+2 to Codex output too.
- [ ] Re-pilot on the same 3 modules from commit `55e0e9a2`; compare resolve rate. Target: ≥60%.
- [ ] Cross-family adversary review per `docs/review-protocol.md` before merge.

## Out of scope

- `--workers N` concurrency coordination via `.pipeline/v2.db` leases (Codex review point #5, deferred to separate issue when we actually parallelize bulk).
- Non-needs_citation resolver classes (that's #344).
- Handling LLMs that refuse to commit to a specific quote (current `unresolvable_reason=all_candidates_failed` path is fine).

## Context

Full pilot writeup in `STATUS.md` (2026-04-23 evening section) and commit `55e0e9a2`. Discussion of why self-verify doesn't work in the conversation leading up to filing.

krisztiankoos:
Closes #356 Part 1 (of 3). Parent: #343.

## Summary

Adds two deterministic filters in \`request_candidates()\` that run after Gemini returns JSON but before any candidate reaches \`validate_candidate()\`:

1. **Allowlist pre-filter** — call \`fetch_citation.allowlist_tier(url)\` and drop anything that returns \`None\`. The LLM still suggests press/news URLs ~30% of the time despite the prompt's hard constraint.
2. **HEAD-check** — new \`head_check(url)\` helper sends HTTP HEAD (5s timeout), falls back to zero-byte range GET on 405. Accepts 2xx/3xx only. Any transport failure = drop.

Both injectable for tests; both threaded through \`resolve_module()\` so the injection surface stays consistent for #356 part 2 (quote-match) and part 3 (Codex fallback).

## Evidence it fixes the pilot problem

Smoke test on the real URLs that caused unresolvable findings in the 2026-04-23 pilot (commit \`55e0e9a2\`):

| URL | Old path | New path |
|---|---|---|
| \`en.wikipedia.org/wiki/Algorithmic_bias\` | pass (real) | pass |
| \`en.wikipedia.org/wiki/Gender_bias_in_machine_learning\` | pass → 404 at fetch → \`http_404\` | **drop at HEAD → never fetched** |
| \`learn.microsoft.com/.../migrate-classic-agents\` | pass → 404 at fetch → \`http_404\` | **drop at HEAD → never fetched** |
| \`reuters.com/article/...\` | pass → \`off_allowlist\` after fetch | **drop at allowlist → never fetched** |

## Test plan

- [x] \`ruff check\` clean
- [x] 23/23 tests pass. 6 new: allowlist filter, 404 filter, 200/404/timeout acceptance, 405→GET-range-0 fallback (asserts Range header).
- [x] Real-world smoke test on 4 pilot URLs (2 live, 2 hallucinated) — all correctly classified.
- [ ] Re-pilot on same 3 modules (follow-up after part 2 lands).

## Review ask

Claude-authored. Routing to Codex for adversary review per \`docs/review-protocol.md\`. Specific questions:

1. Is the HEAD → range-GET fallback on 405 correct? (i.e. are there sites that 405 HEAD and 403 range-GET but 200 a plain GET?) The issue says range-GET is the fallback; I want confirmation this is right.
2. Any missed failure modes in \`head_check\`'s bare-except? I catch everything since "URL is not live enough to cite" is the decision we want in all non-2xx/3xx cases, but I want a second opinion that there's no path where we'd incorrectly drop a live URL.
3. Is there a race condition in the file-write path that my pilot accidentally exercised (two resolvers on the same module clobbering)? If so, should we add a pipeline_v2 lease check here or defer that to a bulk-runner PR? My opinion: defer, since this PR doesn't change the write path.
4. Any regressions in the existing 17 tests that I missed updating?

## Out of scope

- Part 2 (quote-match verification) — separate PR
- Part 3 (Codex fallback) — separate PR
- Concurrency / file locking — separate issue when we implement \`--workers N\`

### DIFF
--- a/scripts/citation_residuals.py
+++ b/scripts/citation_residuals.py
@@ -34,6 +34,7 @@ from citation_backfill import dispatch_gemini, parse_agent_response  # noqa: E40
 HUMAN_REVIEW_DIR = REPO_ROOT / ".pipeline" / "v3" / "human-review"
 DEFAULT_MAX_CANDIDATES = 3
 MIN_SIGNAL_ANCHORS_REQUIRED = 1
+HEAD_CHECK_TIMEOUT_SECONDS = 5.0
 
 
 # ---- IO ------------------------------------------------------------------
@@ -177,14 +178,82 @@ def build_candidate_prompt(finding: dict[str, Any]) -> str:
     )
 
 
+def head_check(
+    url: str,
+    *,
+    timeout: float = HEAD_CHECK_TIMEOUT_SECONDS,
+) -> bool:
+    """Return True if ``url`` resolves to a live 2xx/3xx response.
+
+    Sends an HTTP HEAD first; on 405 Method Not Allowed, falls back to a
+    zero-byte range GET. Any other failure (404, 5xx, timeout, DNS
+    error, connection reset) returns False so the caller can drop the
+    candidate before doing full fetch + anchor matching.
+
+    This is the deterministic gate for URL-existence hallucinations.
+    See GH #356 Part 1. Never raises — any error = "URL is not live
+    enough to cite."
+    """
+    import urllib.error
+    import urllib.request
+
+    def _try(method: str, extra_headers: dict[str, str] | None = None) -> int | None:
+        req = urllib.request.Request(
+            url,
+            method=method,
+            headers={
+                "User-Agent": fetch_citation.DEFAULT_UA,
+                "Accept": fetch_citation.DEFAULT_ACCEPT,
+                **(extra_headers or {}),
+            },
+        )
+        try:
+            with urllib.request.urlopen(req, timeout=timeout) as resp:
+                return int(resp.status)
+        except urllib.error.HTTPError as exc:
+            return int(exc.code)
+        except Exception:  # noqa: BLE001 — any transport failure = dead URL
+            return None
+
+    status = _try("HEAD")
+    if status is None:
+        return False
+    if 200 <= status < 400:
+        return True
+    if status == 405:
+        # Server rejects HEAD but may accept a range-limited GET.
+        retry_status = _try("GET", extra_headers={"Range": "bytes=0-0"})
+        if retry_status is None:
+            return False
+        return 200 <= retry_status < 400
+    return False
+
+
 def request_candidates(
     finding: dict[str, Any],
     *,
     dispatcher=dispatch_gemini,
+    allowlist_tier=fetch_citation.allowlist_tier,
+    head_checker=None,
 ) -> list[dict[str, Any]]:
     """Ask the LLM for candidate URLs. Returns a list of dicts with
     ``url``, ``tier``, ``why``. Empty list on any failure (the caller
-    marks the finding unresolvable)."""
+    marks the finding unresolvable).
+
+    Pre-filters hallucinations deterministically:
+      1. URL must be on ``docs/citation-trusted-domains.yaml`` allowlist.
+         LLMs reliably suggest off-allowlist URLs despite the prompt's
+         hard constraint — this filter is advisory's enforcement.
+      2. URL must respond to HEAD (or range-GET fallback) with a 2xx/3xx.
+         Eliminates the "LLM confidently fabricated a 404 URL" class
+         before downstream fetch + anchor work.
+
+    Both filters are injectable for testing. ``head_checker`` defaults
+    to the module-level ``head_check`` but is resolved at call time so
+    tests can swap it in via parameter.
+    """
+    if head_checker is None:
+        head_checker = head_check
     prompt = build_candidate_prompt(finding)
     ok, raw = dispatcher(prompt)
     if not ok:
@@ -202,6 +271,10 @@ def request_candidates(
             continue
         url = str(c.get("url") or "").strip()
         if not url.startswith(("http://", "https://")):
+            continue
+        if not allowlist_tier(url):
+            continue
+        if not head_checker(url):
             continue
         valid.append(
             {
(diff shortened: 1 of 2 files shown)


krisztiankoos:
Addressed both nits from Codex review in `19e4ab05`:

### Nit #1 — Range rejection escalation

`head_check` now escalates: HEAD → range-GET → plain-GET. The third step only fires when range-GET fails with a Range-specific status (403 / 416 / 501), catching WAF/CDN setups that reject Range but serve plain GETs. Non-Range failures (e.g. 404) still short-circuit, so genuinely dead URLs don't waste extra round-trips.

### Nit #2 — Narrow the bare except

Replaced `except Exception` with an explicit tuple: `URLError, HTTPException, socket.timeout, socket.gaierror, ConnectionError, TimeoutError`. Unexpected exceptions now bubble up instead of becoming silent False returns.

### Tests
26/26 pass (3 new):
- range→plain-GET escalation sequence asserted
- no plain-GET retry after non-Range failure (404)
- `RuntimeError` bubbles up instead of being swallowed

Smoke tests against real URLs unchanged. Ready for re-review or merge.

Issue #2, opened 2025-12-02
Part 1: Cluster Architecture, Installation & Configuration (25%)


krisztiankoos:
## Overview
Second largest exam domain (25%). Understanding how K8s works and extending it.

## Modules (Updated for CKA 2025)

- [ ] **1.1 Control Plane Deep-Dive** - API server, scheduler, controller-manager, kubelet, what each does
- [ ] **1.2 Extension Interfaces** - CNI, CSI, CRI - how Kubernetes is extended
- [ ] **1.3 Helm** - Package management, charts, releases, values files 🆕
- [ ] **1.4 Kustomize** - Configuration management, overlays, patches 🆕
- [ ] **1.5 CRDs & Operators** - Custom Resource Definitions, extending the API 🆕
- [ ] **1.6 RBAC** - Roles, ClusterRoles, Bindings, ServiceAccounts
- [ ] **1.7 kubeadm Basics** - Cluster bootstrap, node join (upgrades deprioritized)

## Exam Weight
**25%** - Approximately 4 questions

## Key Skills (2025)
- Deploy applications using Helm
- Manage configurations with Kustomize
- Understand and create CRDs
- Configure RBAC rules
- Understand extension interfaces (CNI, CSI, CRI)

## What Changed in 2025
| Added | Removed/Deprioritized |
|-------|----------------------|
| Helm | etcd backup/restore |
| Kustomize | Cluster version upgrades |
| CRDs & Operators | Infrastructure provisioning |
| Extension interfaces focus | |

## Common Gotchas
- Helm values precedence (--set vs -f)
- Kustomize base vs overlay confusion
- CRD scope (namespaced vs cluster-wide)
- RBAC verb confusion (get vs list vs watch)

---
⚠️ **Curriculum Version**: CKA 2025 (effective Feb 18, 2025)
📋 **Source**: [CNCF Curriculum](https://github.com/cncf/curriculum)

Issue #1, opened 2025-12-02
Part 0: Environment & Exam Technique


krisztiankoos:
## Overview
Setup and exam technique modules - the foundation for everything else.

## Modules

- [ ] **0.1 Cluster Setup** - kubeadm multi-node cluster (1 CP + 2 workers)
- [ ] **0.2 Shell Mastery** - aliases, autocomplete, .bashrc/.zshrc optimization
- [ ] **0.3 Vim for YAML** - minimal .vimrc, essential motions for YAML editing
- [ ] **0.4 kubernetes.io Navigation** - where to find things fast during exam
- [ ] **0.5 Exam Strategy: Three-Pass Method** - Quick → Medium → Complex workflow

## Priority
🔴 **Critical** - Must complete first. All other modules depend on having a working cluster and efficient environment.

## The Three-Pass Method (Module 0.5)

The core exam strategy taught throughout KubeDojo:

```
Pass 1: Quick Wins
└── Scan all 16 questions
└── Do all tasks you can complete in 1-3 minutes
└── Bank time, build confidence

Pass 2: Medium Tasks  
└── Return to 4-6 minute tasks
└── Steady progress, use standard time

Pass 3: Complex Tasks
└── Troubleshooting, multi-step problems
└── Use ALL remaining time
└── No artificial time pressure
```

## Exam Relevance
Not directly tested, but **determines success**. 

- Efficient environment setup saves 5-10 minutes
- Three-pass strategy maximizes score under time pressure
- Shell mastery = faster execution on every question

---
⚠️ **Curriculum Version**: CKA 2025 (effective Feb 18, 2025)
📋 **Source**: [CNCF Curriculum](https://github.com/cncf/curriculum)

PR #350, opened 2026-04-23
content(cloud): v4 batch-c — 78 modules citation backfill 1.5 → 5.0

krisztiankoos:
## Summary

- Overnight cloud-track run of `pipeline_v4_batch` (2026-04-22 21:35 → 2026-04-23 07:16 UTC, ~9h40m)
- 78 modules processed, all reached `stage_reached: DONE` at `score_after: 5.0` with `gaps_after: []`
- Outcome split: 64 `residuals_filed` (queued for citation triage, but already at 5.0 quality), 14 `clean`
- Score delta: **+262.5 total, avg +3.365/module** (baseline ~1.5 → 5.0)

Matches the pattern of batch A (#345) and batch B (#347), but on the cloud track instead of ai-ml-engineering.

## What's in the diff

- **82 module `.md` files** modified — citation markers added, Gate A overstatement softenings applied, `## Sources` section appended with authoritative links (kubernetes.io, docs.aws.amazon.com, cloud.google.com, official vendor postmortems, etc.)
- **70 citation seed JSONs** added under `docs/citation-seeds/` — the per-module research artifacts consumed by the pipeline; committing these keeps them auditable and reusable for re-runs
- The earlier commit `12141a45` is the 8.2 / 8.3 / 8.5 slice that landed yesterday before the dispatch crash; `f1754f0d` is the overnight remainder

## Run characteristics

- Zero hangs, zero timeouts, zero orphan dispatches
- No `switching to subscription` events in the log — the API-key tier held throughout (the auto-fallback PR is insurance, not load-bearing for this run)

## Test plan

- [x] Pipeline internal recheck: every module final-scored at 5.0
- [ ] Spot-check 2–3 module diffs in review (citations point to real, relevant URLs)
- [ ] `npm run build` — 0 errors, 0 warnings
- [ ] `python scripts/check_site_health.py` — 0 errors

## Residuals (non-blocking)

64 modules have unresolved citation findings queued in `.pipeline/v3/human-review/` — these are tracked by #341 / #343 / #344 (citation-residuals auto-resolver epic) and do not block merge. Per #342 discussion, the `needs_human` outcome label is a misnomer — these modules already cleared the 4/5 rubric gate at 5.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #359, opened 2026-04-23
feat(#351): post-batch audit for modules marked clean but missing `## Sources`

### LINKED ISSUE #351: citation_v3: post-batch audit for modules with no `## Sources` section
## Context

During PR #350 (batch-c cloud citation backfill, 78 modules), Codex review surfaced that `citation_v3` silently left **2 of 78 modules with no `## Sources` section at all**:

- `src/content/docs/cloud/gcp-essentials/module-2.7-cloud-run.md`
- `src/content/docs/cloud/aks-deep-dive/module-7.5-aks-fleet-manager.md`

Both were marked `outcome: "clean"` by the pipeline (no residual findings filed) and cleared the 4/5 rubric gate at score 5.0 — yet the actual citation requirement was unmet. The manual Codex reviewer caught it; our own pipeline did not.

Fix was applied manually in commit `84f56a38`.

## Root cause (hypothesis)

`citation_v3` has two code paths:
1. Module already has a `## Sources` section → append/amend in place.
2. Module lacks a `## Sources` section → append a new one.

One of these paths appears to have a silent-skip branch when the module already has inline `[text](url)` links but no trailing header, or when a template-shaped section exists but is gated out.

## Proposed fix

Two parts:

**A. Post-batch CI audit (cheap, fast to ship).** After `pipeline_v4_batch` finishes, run:

```bash
for module in $(batch_scope); do
  if ! grep -q "^## Sources" "$module"; then
    echo "MISSING SOURCES: $module"
  fi
done
```

and fail the batch run if any module has `outcome: clean` yet no `## Sources`. A 10-line audit script, run automatically.

**B. Root-cause fix in `citation_v3`.** Add a hard post-condition assert at the end of the citation stage: if the module is marked `clean`, the written file MUST contain `^## Sources`. Raise otherwise. This turns a silent failure into a loud one.

## Out of scope

This does not address incomplete `## Sources` (present but thin) — only the zero-sources case.

## Reference

- PR #350 Codex review comment flagging this.
- Handoff doc `STATUS.md` 2026-04-23 entry, pipeline-gap #1.

krisztiankoos:
Closes #351.

## Summary

Adds \`audit_missing_sources(results, repo_root)\` to \`scripts/pipeline_v4_batch.py\`. Walks the batch result list, and for any entry with \`outcome=clean\` checks that the module file on disk contains \`\\n## Sources\\n\`. Wires it into \`main()\` — violations print to stderr and force exit code 1.

## Why

PR #350 (batch-c cloud) silently shipped **2 of 78 modules with outcome=clean but no Sources section at all** (\`module-2.7-cloud-run\`, \`module-7.5-aks-fleet-manager\`). Codex review caught them after the fact; our own pipeline said they were clean. This audit turns the silent-failure class into a loud one.

## Scope / trade-offs

- Matches \`\\n## Sources\\n\` (anchored on start-of-line) to avoid false-positives from inline substrings.
- Skipped under \`--dry-run\` since dry runs don't write modules.
- Flags \"clean but module file missing\" as a violation — if the pipeline claims clean, the file should exist.
- \`run_batch\` summary now includes \`results\` so the audit has something to walk. Additive — existing callers unaffected.

## Test plan

- [x] 7 new unit tests: clean+with-sources passes, clean+no-sources flagged, non-clean outcomes skipped, missing module file flagged, inline substring rejected, CLI integration exit 1, \`--dry-run\` skips.
- [x] Full suite: 390 passed, 1 skipped.
- [x] \`ruff check\` clean.

## Review ask

Claude-authored. Cross-family Codex review per \`docs/review-protocol.md\`. Specific questions:

1. Is the header-detection pattern (\`\\n## Sources\\n\` + \`text.startswith(\"## Sources\\n\")\`) tight enough? Any Markdown edge case where a legitimate Sources section wouldn't match (e.g. CRLF line endings, trailing whitespace on the header line)?
2. Is wiring into \`main()\` the right integration point? Alternative: expose it as a standalone CLI subcommand / separate script that CI calls. My call: inline, because bulk runs should fail fast and operators already watch \`main()\`'s exit code.
3. Should \`--dry-run\` enabling the audit with a warning-only mode be a follow-up? My call: no, dry-run is genuinely incompatible with the audit (no module edits occur).
4. Any regressions in the existing 16 \`test_pipeline_v4_batch.py\` tests I missed updating?

### DIFF
--- a/scripts/pipeline_v4_batch.py
+++ b/scripts/pipeline_v4_batch.py
@@ -38,6 +38,7 @@ import argparse
 import datetime as dt
 import json
 import os
+import re
 import sqlite3
 import sys
 import threading
@@ -85,7 +86,10 @@ CREATE INDEX IF NOT EXISTS idx_v4_batch_expires ON v4_batch_leases(expires_at);
 OUTCOME_SKIPPED_LOCKED = "skipped_locked"
 OUTCOME_ERROR = "error"
 OUTCOME_CLEAN = "clean"
-SOURCES_HEADER_PATTERN = "\n## Sources\n"
+# ATX H2 named "Sources", anchored to start-of-line, trailing whitespace OK.
+# Mirrors the convention used in other repo checkers; tolerates
+# `## Sources  \n` and EOF-without-final-newline per Codex #359 review.
+SOURCES_HEADER_RE = re.compile(r"^##\s+Sources\s*$", re.MULTILINE)
 
 _INIT_LOCK = threading.Lock()
 _INITIALIZED_DBS: set[str] = set()
@@ -349,10 +353,12 @@ def audit_missing_sources(
             violations.append(module_key)
             continue
         text = module_path.read_text(encoding="utf-8")
-        # Match a ## Sources header at start-of-line (not inside a code
-        # block or inline prose). Anchoring on a leading newline catches
-        # both mid-file and end-of-file cases.
-        if SOURCES_HEADER_PATTERN not in text and not text.startswith("## Sources\n"):
+        # MULTILINE regex against an ATX H2 named "Sources" with optional
+        # trailing whitespace. Start-of-line anchor avoids false-positives
+        # from the string appearing mid-prose; trailing \s* tolerates the
+        # `## Sources  ` (trailing spaces) and EOF-without-newline cases
+        # Codex flagged on PR #359.
+        if not SOURCES_HEADER_RE.search(text):
             violations.append(module_key)
     return violations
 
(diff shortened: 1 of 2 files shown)


krisztiankoos:
Addressed both must-fix findings in `1fad9f3e`:

### Must-fix #1 — `--skip-citation` false-positive

Main() now skips the audit under `--skip-citation` (in addition to `--dry-run`). Rationale: pipeline_v4 explicitly bypasses Stage 4 in that mode and can legitimately return `outcome=clean` for modules that lifted from no-quiz / no-exercise work alone; flagging them would turn a supported mode into a failure. New regression test: `test_cli_skip_citation_skips_missing_sources_audit`.

### Must-fix #2 — header regex too strict

Replaced the exact-string match with a MULTILINE regex `^##\s+Sources\s*$`, mirroring the convention in the repo's other checkers. Now accepts trailing whitespace (`## Sources  `) and EOF-without-newline cases. Two regression tests: `test_audit_accepts_sources_header_with_trailing_whitespace` + `test_audit_accepts_sources_header_at_eof_without_trailing_newline`.

Dropped the now-unused `SOURCES_HEADER_PATTERN` string constant.

26/26 tests pass (was 23). ruff clean.

krisztiankoos:
Codex re-review (task pr-359-rereview): **APPROVE**. No findings, no nits.

> I verified both prior must-fix items are actually fixed on `1fad9f3e`. [...] Existing coverage at `tests/test_pipeline_v4_batch.py:492` still rejects inline prose false positives.

Merging.

PR #360, opened 2026-04-23
feat(#356 part 2): quote-match verification rescues anchorless findings

### LINKED ISSUE #356: #343 phase-1.5: HEAD pre-check + quote-match verification + Codex fallback (URL hallucination mitigation)

krisztiankoos:
Closes #356 Part 2 (of 3). Parent: #343. Builds on Part 1 (#357, merged).

## Summary

Extends the LLM candidate prompt to require \`expected_quote\` — a specific sentence or phrase from the page the LLM claims to cite. \`validate_candidate()\` fetches the page and substring-matches the quote (case-insensitive, whitespace-normalized, ≥12-char match, prefix-tolerance for LLM drift) before accepting.

This catches the **second** hallucination class that Part 1 didn't: URL exists on allowlist, page returns 200, but the page doesn't actually discuss the claim. Also rescues the **anchorless findings class** (signals like \`named_incident\` / \`attribution\` alone) that phase-1 had to reject outright.

## Verification precedence

\`\`\`
no anchors AND no quote  → reject pre-fetch (no_verification_signal)
anchors present, matched → ok, verified_via=anchors
quote present, matched   → ok, verified_via=expected_quote
neither matched          → reject (quote_not_in_page or no_anchor_match)
\`\`\`

When both anchors and quote are present and both match, report \`verified_via=anchors\` (cheaper, more trusted signal).

## Drift tolerance

\`quote_present_in_text()\`:
1. Normalize whitespace (\`\\s+\` → single space) and lowercase on both sides.
2. Require ≥12-char minimum (rejects LLM hedging like \"see there\").
3. Full-quote substring match wins.
4. If full miss, accept a word-boundary-aligned prefix match — so \"Amazon scrapped its AI recruiting tool after discovering bias\" still accepts when the page says \"Amazon scrapped its AI recruiting tool after finding issues\" (tail changed).

## Expected impact

Re-pilot on the 2026-04-23 modules should now resolve the \`named_incident\` / \`attribution\` findings that went to \`unresolvable\` under phase-1 (1 in fine-tuning-llms, 1 in cloud-ai-services, 1 in multi-account). Projected resolve rate: 60-75% (vs. 20% on the original pilot).

## Test plan

- [x] 8 new tests:
  - 4 on \`quote_present_in_text\`: happy path, too-short reject, tail tolerance, unrelated reject
  - 3 on \`validate_candidate\`: anchorless+quote accepts, quote-not-on-page rejects, both-present reports anchors
  - 1 on \`request_candidates\`: \`expected_quote\` survives JSON parse
- [x] 34/34 total pass (was 26).
- [x] \`ruff check\` clean.
- [ ] Re-pilot against 2026-04-23 modules (follow-up after merge).

## Review ask

Claude-authored. Cross-family Codex review per \`docs/review-protocol.md\`. Specific questions:

1. \`MIN_QUOTE_MATCH_LENGTH = 12\` — too lenient (false-positives on generic phrases) or too strict (misses short valid quotes)? Any data-driven way to choose this besides hand-waving?
2. The prefix-match fallback (\`prefix = norm_quote[:max(12, len/2)]\`, rounded to word boundary): is there a case where a legitimate prefix match produces a false positive? E.g. the LLM starts with a common phrase like \"According to the article,\" that's on many pages.
3. When both anchors AND quote match, I report \`verified_via=anchors\`. Is that the right precedence, or should the more-specific signal (quote) win?
4. \`request_candidates\` now also accepts candidates with empty \`expected_quote\`. Those will then be rejected at \`validate_candidate\` if no anchors. Should \`request_candidates\` drop them earlier? My call: no, the layered rejection gives better diagnostics in \`attempts[]\`. Confirm.
5. Any other bugs / missed edge cases.

### DIFF
--- a/scripts/citation_residuals.py
+++ b/scripts/citation_residuals.py
@@ -328,14 +328,19 @@ def normalize_quote(text: str) -> str:
 
 
 def quote_present_in_text(quote: str, body: str) -> bool:
-    """Check whether ``quote`` (or a long-enough prefix) is a substring
-    of ``body`` after whitespace + case normalization.
+    """Check whether ``quote`` is present in ``body`` after whitespace +
+    case normalization.
 
-    Accepts matches of at least ``MIN_QUOTE_MATCH_LENGTH`` characters
-    so minor LLM transcription drift (trailing phrase, missing word)
-    doesn't reject an otherwise-correct citation. Returns False when
-    the quote is too short to be meaningful — that's the LLM hedging,
-    not evidence.
+    Primary path: the full normalized quote must be a substring of the
+    body. Drift-tolerance fallback: BOTH a start window AND an end
+    window (each ≥ ``MIN_QUOTE_MATCH_LENGTH`` chars, word-boundary-
+    aligned) must appear in the body in order. This rejects generic
+    lead-ins like "According to the article..." where only the prefix
+    matches but the claim-specific tail differs (#360 Codex review,
+    must-fix #2).
+
+    Returns False when the normalized quote is too short to be
+    meaningful — that's the LLM hedging, not evidence.
     """
     norm_quote = normalize_quote(quote)
     if len(norm_quote) < MIN_QUOTE_MATCH_LENGTH:
@@ -343,14 +348,44 @@ def quote_present_in_text(quote: str, body: str) -> bool:
     norm_body = normalize_quote(body)
     if norm_quote in norm_body:
         return True
-    # Tolerate LLM drift on the tail: accept if the first
-    # MIN_QUOTE_MATCH_LENGTH-char prefix matches, rounded up to a word
-    # boundary so we don't mid-word match accidentally.
-    prefix = norm_quote[:max(MIN_QUOTE_MATCH_LENGTH, len(norm_quote) // 2)]
-    prefix = prefix.rsplit(" ", 1)[0] if " " in prefix else prefix
-    if len(prefix) >= MIN_QUOTE_MATCH_LENGTH and prefix in norm_body:
-        return True
-    return False
+    # Drift-tolerance fallback: require the quote's start AND end
+    # windows to both appear in the body, in order. The quote must be
+    # roughly 2× MIN + small gap so the start and end windows don't
+    # overlap.
+    if len(norm_quote) < 2 * MIN_QUOTE_MATCH_LENGTH + 4:
+        return False
+    start_window = _trim_to_trailing_word_boundary(
+        norm_quote[: MIN_QUOTE_MATCH_LENGTH + 4]
+    )
+    end_window = _trim_to_leading_word_boundary(
+        norm_quote[-(MIN_QUOTE_MATCH_LENGTH + 4):]
+    )
+    if (
+        len(start_window) < MIN_QUOTE_MATCH_LENGTH
+        or len(end_window) < MIN_QUOTE_MATCH_LENGTH
+    ):
+        return False
+    start_idx = norm_body.find(start_window)
+    if start_idx < 0:
+        return False
+    return end_window in norm_body[start_idx + len(start_window):]
+
+
+def _trim_to_trailing_word_boundary(text: str) -> str:
+    """Drop a trailing partial word so a window ends on a clean
+    whitespace boundary. Returns the original string when no space
+    exists to trim against."""
+    if " " not in text:
+        return text
+    return text.rsplit(" ", 1)[0]
+
+
+def _trim_to_leading_word_boundary(text: str) -> str:
+    """Mirror of _trim_to_trailing_word_boundary for the start of a
+    window: drop a leading partial word."""
+    if " " not in text:
+        return text
+    return text.split(" ", 1)[1]
 
 
 # ---- validation ----------------------------------------------------------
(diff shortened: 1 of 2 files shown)


krisztiankoos:
Addressed both Codex must-fix findings in `3404ab8b`:

### Must-fix #1 — anchors authoritative

Old logic accepted the candidate if EITHER anchors OR quote matched, so a generic quote could override a contradictory anchor (your example: claim says 2018, page says 2017, but quote 'Amazon scrapped its AI tool' still matched → ok=True). Fixed: when the finding has extractable anchors, those are authoritative. Anchor mismatch → `no_anchor_match` reject, regardless of quote result. Quote-only acceptance is limited to anchorless findings. New regression test `test_validate_candidate_rejects_anchor_mismatch_even_when_quote_matches` uses your exact example.

### Must-fix #2 — stricter drift tolerance

Replaced prefix-only fallback with a start-window AND end-window match (both ≥ MIN_QUOTE_MATCH_LENGTH chars, word-boundary-aligned, in order). Rejects your adversarial example (generic 'According to the article...' lead-in with the claim's subject swapped from Amazon to Kubernetes). Two new regression tests: `test_quote_present_rejects_end_changed` and `test_quote_present_rejects_generic_lead_in_with_different_subject` (your exact example).

The existing drift-tolerance test was renamed `test_quote_present_tolerates_minor_internal_drift` — bookend-match still allows drift IN THE MIDDLE of the quote, just not at the tail.

37/37 tests pass. Ready for re-review.

PR #366, opened 2026-04-24
docs(#343): residuals classification audit — 195 findings across 92 files

krisztiankoos:
## Summary

- Classifies all 195 \`needs_citation\` findings across 92 residuals queue files (\`.pipeline/v3/human-review/\`) into four actionable buckets: Sourceable (57%), Pedagogical fiction (33%), Hallucinated fact (0.5%), Ambiguous (9%).
- Drives scope decisions across the citation-residuals epics: #343 (core resolver) targets Category A; #344 (phase-2 overstated/off-topic) owns Category B; new issue #365 tracks the single Category C.
- Authored by Codex via headless dispatch (task \`audit-residuals-classification\`); committed from the primary repo side per the sandbox-commit pattern (Codex sandbox blocks \`.git/worktrees/*/index.lock\`).

## Context

Overnight push 2026-04-24: the 3-module resolver pilot returned 1/4 resolved. Diagnosis (captured in STATUS.md) reframed the failure — the resolver was correctly refusing to polish fabrications (war-story openers, hallucinated future dates). The question that fell out: \"What fraction of the 64 batch-c residuals modules are like that?\" This audit answers it.

## Impact on the open epics

- **#343 (core resolver)**: the real addressable population is ~57% (Category A), not the full queue. Before bulk-running phase-2 at \`--workers 3\`, resolver description formatting must be fixed — see the new issue #364.
- **#344 (phase-2 overstated/off-topic)**: Category B (65 findings) is the workload. These are narrative devices, not citation-resolver misses.
- **#365 (content bug — CloudWatch/EC2)**: the single Category C finding. Filed separately; scope is content-fix, not pipeline.

## Related

- Closes the audit AC of #343.
- Unblocks phase-2 bulk planning once #364 lands.
- Cross-references #344, #365, #341 (epic).

## Test plan

- [ ] Cross-family review per \`docs/review-protocol.md\` (Codex authored → Claude or Gemini reviews; prefer Gemini for content classification review).
- [ ] Verify audit counts match source files: \`.venv/bin/python -c 'import json,glob; t=sum(len(json.load(open(f)).get(\"queued_findings\",{}).get(\"needs_citation\",[])) for f in glob.glob(\".pipeline/v3/human-review/*.json\")); print(t)'\` — expect 195.
- [ ] Spot-check 3 Category B assignments to confirm the narrative-fiction label.
- [ ] Build green: \`npm run build\` (docs-only change, no Starlight impact expected).

Issue #364, opened 2026-04-24
citation-residuals: Sources-line description uses finding excerpt instead of describing the cited source


krisztiankoos:
Discovered during overnight push — the multi-account pilot resolve produced this Sources line:

\`\`\`
- [docs.aws.amazon.com: benefits of using multiple aws accounts.html](https://docs.aws.amazon.com/whitepapers/latest/organizing-your-aws-environment/benefits-of-using-multiple-aws-accounts.html) — The root cause of this catastrophic failure was not the load test itself, nor was it the junior developer's actions.
\`\`\`

The URL is correct. The trailing description, however, is the **original claim text from the module narrative** (the war-story opening), not a description of what the AWS whitepaper says. To readers it reads as though the whitepaper authored that claim — confusing and potentially misleading.

## Root cause

\`scripts/citation_residuals.py:537-547\` — \`_summarize_finding()\` uses the first sentence of the finding's \`excerpt\` field (truncated to 160 chars) as the note:

\`\`\`python
def _summarize_finding(finding: dict[str, Any]) -> str:
    excerpt = (finding.get(\"excerpt\") or \"\").strip()
    # First sentence of excerpt, capped at 160 chars — enough for context
    # without duplicating the whole paragraph.
    first_sentence = re.split(r\"(?<=[.!?])\\s+\", excerpt, maxsplit=1)[0]
    ...
\`\`\`

But \`excerpt\` is the **claim being cited**, not a description of what the source actually says. The comment (\"enough for context\") reveals the original intent — the implementer confused \"context\" (what this source supports) with \"description\" (what this source says).

## Why this blocks #343 phase-2 bulk

Every successful resolve currently produces a garbage Sources description. Running the planned bulk (64 modules × ~2–3 resolved findings each) would ship ~100+ broken Sources lines into the curriculum. Must be fixed before \`--workers 3\` bulk.

## Proposed fix options

1. **Title-based description** (simplest): drop the description altogether and use a richer \`_derive_title_from_url\` — e.g., \"AWS Whitepaper: benefits of using multiple AWS accounts\". Risk: titles become less informative.
2. **LLM-generated one-liner** (preferred): during resolve, ask the LLM for a short one-sentence description of what the page documents (\"AWS whitepaper explaining blast-radius, IAM boundary, and cost-attribution benefits of multi-account architectures\"). One extra call per resolve; cached alongside the URL. Structured output.
[...]

PR #367, opened 2026-04-24
feat(#343): --limit-modules N for gradual pilot rollout

krisztiankoos:
## Summary

- Adds \`--limit-modules N\` to \`citation_residuals.py resolve --all\` so operators can cap a run at N modules without editing the script — the pilot-N knob the 2026-04-23 postmortem asked for.
- Applies the cap AFTER the empty-queue filter, so \`N=10\` means 10 modules with real findings, not 10 scanned files.
- No interaction with #363's per-module lock: \`--limit-modules\` is an outer filter on the loop; \`module_lock.acquire_module_lock\` still wraps each \`resolve_module\` call.
- Refs #343, #363, \`feedback_batch_worker_cap.md\`.

## Usage

\`\`\`bash
# Pilot-10 on fresh batch-c residuals:
python scripts/citation_residuals.py resolve --all --worker-id pilot-2 --limit-modules 10

# Ignored with a single module_key (the single-target path is already N=1):
python scripts/citation_residuals.py resolve ai/demo/module-1.0-test --limit-modules 5
\`\`\`

Prints the cap in effect:

\`\`\`
--limit-modules 10: processing 10 of 64 useful modules
\`\`\`

## Why

Phase-2 bulk at \`--workers 3\` across 64 batch-c residuals modules is the target after #363 landed. Running bulk blind is risky — the audit (PR #366) shows only ~57% of findings are Category A (sourceable), with the rest being pedagogical fiction or ambiguous. A gradual \`--limit-modules 10\` pilot on a fresh sample sets a calibrated resolve-rate before the full run, and keeps the operator loop tight.

Also surfaces early whatever #364 (Sources-line description bug) would produce at scale, before the bulk commits ship.

## Test plan

- [x] 2 new tests: cap applied with \`--all\`; ignored when \`module_key\` given.
- [x] 39/39 existing \`test_citation_residuals.py\` pass (including the lock / canonical-key tests from #363).
- [x] Ruff clean.
- [ ] Cross-family review per \`docs/review-protocol.md\` — prefer Codex (more rigorous on Claude-authored code batches per #350 data point).
- [ ] Smoke: \`python scripts/citation_residuals.py resolve --all --limit-modules 1 --dry-run\` on a real residuals snapshot processes exactly one useful module.

### DIFF
--- a/scripts/citation_residuals.py
+++ b/scripts/citation_residuals.py
@@ -817,6 +817,20 @@ def main(argv: list[str] | None = None) -> int:
     )
     p_resolve.add_argument("--all", action="store_true", help="Process every queue file")
     p_resolve.add_argument(
+        "--limit-modules",
+        type=int,
+        default=None,
+        metavar="N",
+        help=(
+            "Cap the number of modules processed in an --all run to N. Useful "
+            "for gradual rollout / pilot runs before a full bulk batch — see "
+            "the 2026-04-23 postmortem. Applied AFTER the empty-queue filter, "
+            "so N=10 means 10 modules with actual needs_citation findings, "
+            "not 10 scanned files. Has no effect when a specific module_key is "
+            "given instead of --all."
+        ),
+    )
+    p_resolve.add_argument(
         "--dry-run",
         action="store_true",
         help="Propose resolutions but do not write to modules or queue JSON",
@@ -887,6 +901,14 @@ def main(argv: list[str] | None = None) -> int:
         if not useful_targets:
             print("No residuals with needs_citation findings.")
             return 0
+
+        if args.all and args.limit_modules is not None and args.limit_modules > 0:
+            total_before = len(useful_targets)
+            useful_targets = useful_targets[: args.limit_modules]
+            print(
+                f"--limit-modules {args.limit_modules}: "
+                f"processing {len(useful_targets)} of {total_before} useful modules"
+            )
 
         totals = {
             "considered": 0,
--- a/scripts/citation_residuals.py
+++ b/scripts/citation_residuals.py
@@ -816,18 +816,36 @@ def main(argv: list[str] | None = None) -> int:
         help="Module key or slug. Use --all instead to process every queued file.",
     )
     p_resolve.add_argument("--all", action="store_true", help="Process every queue file")
+
+    def _positive_int(raw: str) -> int:
+        try:
+            value = int(raw)
+        except ValueError as exc:
+            raise argparse.ArgumentTypeError(
+                f"--limit-modules must be a positive integer, got {raw!r}"
+            ) from exc
+        if value <= 0:
+            raise argparse.ArgumentTypeError(
+                f"--limit-modules must be >= 1 (got {value}); "
+                "silently treating 0 or negative as 'no limit' would let a "
+                "typo unleash the full bulk run"
+            )
+        return value
+
     p_resolve.add_argument(
         "--limit-modules",
-        type=int,
+        type=_positive_int,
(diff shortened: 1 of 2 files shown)


krisztiankoos:
NEEDS CHANGES

FINDING: `--limit-modules` treats `0` and negative values as "no limit", so a typo can run the full bulk queue  
FILE:LINE: `scripts/citation_residuals.py:905`  
CURRENT CODE:
```python
if args.all and args.limit_modules is not None and args.limit_modules > 0:
    total_before = len(useful_targets)
    useful_targets = useful_targets[: args.limit_modules]
```
WHY: `--limit-modules 0` and `--limit-modules -1` skip this guard and fall through to processing every useful module. I reproduced that on the PR worktree: both values ran all 3 queued modules. For a rollout cap, silently turning an invalid limit into a full bulk run is the dangerous failure mode, and it also makes the help text inaccurate because the flag has an undocumented extra no-op case.  
FIX: Reject non-positive values up front with argparse validation or an explicit `parser.error`, and add tests for `0` and negative inputs.

FINDING: the new tests do not actually lock in the documented "apply the cap after the empty-queue filter" behavior  
FILE:LINE: `tests/test_citation_residuals.py:903`  
CURRENT CODE:
```python
for i in range(3):
    _write_queue_file(
        queue_dir / f"ai-demo-module-1.{i}-test.json",
        f"ai/demo/module-1.{i}-test",
        [finding],
    )
```
WHY: every fixture queue file in `test_limit_modules_caps_all_run` has a `needs_citation` entry, so the test still passes if someone slices `targets` before the empty-queue filter. The implementation is correct today, but this PR explicitly documents that ordering as critical and the tests do not defend it.  
FIX: Add a mixed fixture set with empty and non-empty queue files and assert that `--limit-modules 1` processes one useful module after filtering, not the first scanned file.

Checks I ran:
- `../../.venv/bin/python -m pytest tests/test_citation_residuals.py -k 'limit_modules' -q` in the PR worktree: `2 passed`
- `../../.venv/bin/python scripts/citation_residuals.py resolve --help`: help text renders as expected
- Manual repros: `N > total` behaves fine; cap is currently applied after the empty-queue filter; per-module lock ordering looks fine because the cap is an outer slice on `useful_targets` and locking happens per selected module with the canonical key

`gh pr comment 367 --repo kube-dojo/kube-dojo.github.io --body-file -` failed in this sandbox with:
`error connecting to api.github.com`
`check your internet connection or https://githubstatus.com`

krisztiankoos:
Verdict: APPROVE

No remaining must-fix findings.

I verified the branch content at `2de2e57b` rather than relying on the diff summary:
- `scripts/citation_residuals.py:820-848` adds `_positive_int` and wires it into `--limit-modules`.
- `scripts/citation_residuals.py:923-929` applies the cap after `useful_targets` is built.
- `tests/test_citation_residuals.py:982-1062` adds the mixed empty/non-empty ordering test plus the three negative-path argparse tests.

Focused verification on the PR worktree:
- `/Users/krisztiankoos/projects/kubedojo/.venv/bin/pytest -q tests/test_citation_residuals.py::test_limit_modules_applied_after_empty_queue_filter tests/test_citation_residuals.py::test_limit_modules_rejects_zero tests/test_citation_residuals.py::test_limit_modules_rejects_negative tests/test_citation_residuals.py::test_limit_modules_rejects_non_integer`
- Result: `4 passed in 0.12s`

I could not post the PR comment because network is blocked here. Exact `gh` error:
`error connecting to api.github.com`
`check your internet connection or https://githubstatus.com`

Issue #344, opened 2026-04-22
citation-residuals: extend resolver to overstated/off-topic findings


krisztiankoos:
Parent epic: #341
Builds on: #343 (core resolver shipping phase-1 for `needs_citation`)

## Scope

The `queued_findings` JSON has four buckets beyond `needs_citation`:

1. `overstated_unfixed` — claim is too strong ("always", "never", "the best"); auditor wanted to soften it but didn't.
2. `off_topic_unfixed` — paragraph flagged as off-topic; auditor wanted to delete it but didn't.
3. `overstatement_queued` — softening is pre-composed, ready to apply.
4. `off_topic_delete_queued` — deletion is pre-composed, ready to apply.

For the 3 and 4 buckets, the fix is already known — it just needs to be applied to the module. That's a straightforward patch + commit.

For the 1 and 2 buckets, the fix requires an LLM call to propose a softening or deletion, then an apply step.

## Why phase 2

Volume and risk:
- `needs_citation` was 276 of 276 in the first sample — the dominant bucket. Phase 1 takes most of the pain.
- `overstated_unfixed` / `off_topic_unfixed` are content-mutation operations (changing claim strength, deleting paragraphs). Higher review bar than adding a citation marker.
- Queued fixes (3, 4) are pre-composed so in principle safe to apply, but real-world examples should be sampled first.

## Acceptance

- [ ] Resolver's `resolve` subcommand handles all four buckets via a flag: `--fix-overstatements`, `--fix-offtopic`, `--apply-queued`. Default remains `--fix-citations` (phase 1 behavior).
- [ ] Pilot on 2 modules per bucket before bulk.
- [ ] Sampling policy: `--sample N` flag runs resolver on N random findings, emits before/after, then exits — for spot-checks.
- [ ] Gemini review per `docs/review-protocol.md`.

## Out-of-scope

- Coherence / text-level rewrites beyond the scoped overstatement softening.
- Re-running citation_v3 to audit the post-resolver state (possibly a follow-on: re-audit → new findings loop).

Issue #341, opened 2026-04-22
Epic: Citation residuals auto-resolver — scale fact-checking beyond manual review


krisztiankoos:
## Problem

`pipeline_v4`'s Stage 4 (citation_v3) audits each module for fact-level claims (prices, years, named incidents, cited statistics) and when it can't auto-resolve them, files them to `.pipeline/v3/human-review/<module>.json` and marks the module `outcome: "needs_human"` — but the module has already cleared the 4/5 rubric gate at score **5.0** with `gaps_after: []`.

From the first ai-ml-engineering batch (~42 modules as of 2026-04-22 morning):

- 36 modules flagged `needs_human`, 1 genuinely failed, 6 fully `clean`
- **276 total findings queued** across the 36 modules
- Avg **7.7 findings/module**, max 46
- Top signals: `price_usd` (96), `year_reference` (57), `dated_year` (27), `named_incident` (21), `named_company_cap` (21)

Extrapolating to the full below-4 pool (~619 modules as of 2026-04-22): if 2/3 get flagged with ~7.7 findings each ≈ **~3,200–4,100 residual findings** at full #180 completion. Not humanly reviewable.

## Context

- Parent campaign: #180 (Elevate All Modules to 4/5 Quality)
- Driver-update correction: https://github.com/kube-dojo/kube-dojo.github.io/issues/180#issuecomment-4291826907 (explains why Stage 4 is required for non-pre-cited tracks)
- Unblocking fixes that made this scale possible: `2ec9827e`, `57c5f6f9`, `56b0f998`

## Scope

A new `scripts/citation_residuals.py` CLI that reads `.pipeline/v3/human-review/*.json` and, for each queued finding:

1. Generates candidate URLs from the finding's `search_hint` + `excerpt` (via a Gemini dispatch)
2. Validates each candidate via `scripts/fetch_citation.fetch` (tier check + content fetch + link-rot)
3. Picks the best-tier valid URL
4. Injects an inline citation marker in the module `.md` at the flagged line
5. Removes the resolved finding from the queue JSON, updating status to `resolved` with provenance

Unresolvable findings stay in the queue as genuinely-stuck residuals (expected to be a tiny fraction once auto-resolver is tuned).

## Non-goals

- **Not** replacing citation_v3's auditor — the auditor continues to flag; the resolver only consumes the flagged output.
- **Not** re-scoring the module — score already at 5.0; resolver only improves factual density.
[...]

PR #368, opened 2026-04-24
fix(#364): drop finding-excerpt-as-description in Sources line

### LINKED ISSUE #364: citation-residuals: Sources-line description uses finding excerpt instead of describing the cited source

krisztiankoos:
## Summary

Unblocks #343 phase-2 bulk. The resolver's \`_summarize_finding\` took the finding's \`excerpt\` (the module's original claim text) and injected it as the Sources-line description. Readers saw lines like:

\`\`\`
- [docs.aws.amazon.com: benefits...](url) — The root cause of this catastrophic failure was not the load test itself, nor was it the junior developer's actions.
\`\`\`

The URL is valid (AWS whitepaper). The description reads as though the whitepaper authored that claim. Incompatible with the human-curated Sources lines already in the module.

## What this PR does

- Delete \`_summarize_finding\` entirely.
- \`build_source_line\` emits \`- [title](url)\` — no trailing description.
- Strip \`.html\`/\`.htm\` from URL-derived titles so the page reads as a name, not a filename. \`.pdf\` etc. preserved (format signal).
- \`finding\` stays in the signature for call-site compatibility but is documented as intentionally unused.

Before:

\`\`\`
- [docs.aws.amazon.com: benefits of using multiple aws accounts.html](url) — The root cause of this catastrophic failure was not the load test itself.
\`\`\`

After:

\`\`\`
- [docs.aws.amazon.com: benefits of using multiple aws accounts](url)
\`\`\`

## What this PR intentionally does NOT do

- **No LLM one-liner describing the page content.** That is the preferred higher-quality fix but requires either a second dispatch per resolve or a page-body parser. An honest absence beats a misleading description, so ship minimum viable first.
- **Does not re-resolve already-committed garbage descriptions.** A later content sweep or re-resolve will catch those. Out of scope here.

## Tests

- Old \`test_build_source_line_safe_title_and_summary\` (which locked in the bug) replaced by \`test_build_source_line_omits_finding_excerpt\` — a regression guard asserting the output contains neither excerpt text nor a description separator. **If this ever regresses, every phase-2 bulk resolve will re-introduce claim-text-as-description lines.**
- 3 new tests covering escape-bracket safety, explicit-title override, and extension-stripping policy.
- 47/47 tests pass. Ruff clean.

## Refs

- Closes #364.
- Unblocks #343 phase-2 bulk pilot (fresh 10-module sample using the new \`--limit-modules 10\` flag from #367).
- Related: #344 (phase-2 overstated/off-topic scope), #341 (epic).

## Test plan

- [x] 47/47 existing tests pass.
- [x] Smoke: \`build_source_line\` on real vendor URLs (AWS whitepaper, learn.microsoft, cloud.google) matches existing Sources-line convention style, minus description.
- [ ] Cross-family review per \`docs/review-protocol.md\` — Claude-authored, so Codex reviewer preferred.
- [ ] Post-merge: run \`resolve --all --limit-modules 1 --dry-run\` on a real residuals snapshot to confirm a resolved finding produces a clean \`- [title](url)\` line.

### DIFF
--- a/scripts/citation_residuals.py
+++ b/scripts/citation_residuals.py
@@ -514,11 +514,21 @@ def short_id_for(url: str) -> str:
 
 
 def build_source_line(url: str, finding: dict[str, Any], *, title: str | None = None) -> str:
-    """Render one `- [title](url) — note` bullet for the Sources list."""
+    """Render one ``- [title](url)`` bullet for the Sources list.
+
+    ``finding`` is retained in the signature for call-site compatibility
+    but is intentionally unused: earlier versions appended a "description"
+    derived from the finding's excerpt, which was the *claim being cited*
+    rather than a description of what the source actually says (see #364).
+    A URL-derived title is sufficient provenance; richer per-source
+    descriptions — if wanted — should come from the cited page itself
+    (``<title>``, ``<meta description>``) or an LLM summary of the page
+    body, never from the module text.
+    """
+    del finding  # intentionally unused; see docstring
     display_title = title or _derive_title_from_url(url)
-    note = _summarize_finding(finding)
     safe_title = display_title.replace("[", r"\[").replace("]", r"\]")
-    return f"- [{safe_title}]({url}) — {note}"
+    return f"- [{safe_title}]({url})"
 
 
 def _derive_title_from_url(url: str) -> str:
@@ -528,23 +538,17 @@ def _derive_title_from_url(url: str) -> str:
     host = (parsed.hostname or "").removeprefix("www.")
     path_segments = [seg for seg in parsed.path.split("/") if seg]
     tail = path_segments[-1] if path_segments else ""
+    # Strip the document extension (.html/.htm) so the title reads as a
+    # page name rather than a filename. Non-doc extensions (.pdf, .json)
+    # are preserved because they carry meaningful format signal.
+    for ext in (".html", ".htm"):
+        if tail.lower().endswith(ext):
+            tail = tail[: -len(ext)]
+            break
     readable = tail.replace("-", " ").replace("_", " ").strip()
     if host and readable:
         return f"{host}: {readable}"
     return host or url
-
-
-def _summarize_finding(finding: dict[str, Any]) -> str:
-    excerpt = (finding.get("excerpt") or "").strip()
-    # First sentence of excerpt, capped at 160 chars — enough for context
-    # without duplicating the whole paragraph.
-    first_sentence = re.split(r"(?<=[.!?])\s+", excerpt, maxsplit=1)[0]
-    summary = first_sentence[:160].strip()
-    if not summary:
-        return "Supporting citation for the flagged claim."
-    if not summary.endswith((".", "!", "?")):
-        summary += "."
-    return summary
 
 
 def inject_source(module_text: str, source_line: str) -> tuple[str, bool]:
--- a/tests/test_citation_residuals.py
+++ b/tests/test_citation_residuals.py
@@ -871,7 +871,15 @@ def test_resolve_module_marks_unresolvable_when_no_candidates(
 # ---- source_line rendering -----------------------------------------------
 
 
-def test_build_source_line_safe_title_and_summary() -> None:
+def test_build_source_line_omits_finding_excerpt() -> None:
+    """Regression guard for #364. The Sources-line description MUST NOT
+    derive from the finding's excerpt — the excerpt is the claim being
+    cited, not a description of what the source documents. Earlier
+    resolver versions shipped the excerpt verbatim as the description,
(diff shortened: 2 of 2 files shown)


krisztiankoos:
APPROVE

Must-fix findings: none.

Notes:
- `scripts/citation_residuals.py:516-551` looks correct for the bug in #364. The new output is plainer than the hand-curated style in `src/content/docs/cloud/advanced-operations/module-8.1-multi-account.md:1114-1125`, but the repo already contains some title-only `Sources` bullets, so I don’t think that inconsistency is blocking.
- `.html/.htm` stripping in `scripts/citation_residuals.py:541-547` is the right call. Query strings and anchors are ignored correctly by `urlparse()`, trailing-slash URLs still derive from the last path segment, bare-host URLs fall back to the host, and non-doc extensions like `.pdf` remain intact.
- Keeping `finding` and explicitly `del`eting it at `scripts/citation_residuals.py:519-529` is acceptable for a minimal-risk fix. Simplifying the signature would be cleanup, not a correctness issue.
- `tests/test_citation_residuals.py:874-899` does guard the claimed regression. A reintroduction of excerpt-as-description would fail both on the missing ` — ` assertion and on the explicit excerpt substring checks.
- I also smoke-tested the reviewed commit’s resolve path in an extracted `/tmp` tree by monkeypatching candidate discovery/validation and invoking `main(['resolve', '--all', '--limit-modules', '1', '--dry-run'])`. The captured emitted line was `- [docs.aws.amazon.com: aft overview](https://docs.aws.amazon.com/controltower/latest/userguide/aft-overview.html)`, which matches the intended title-only behavior.

PR #711, opened 2026-05-01
feat(status): add public quality progress page (#391)

### LINKED ISSUE #391: Public-facing status page: rewrite lifecycle visibility for learners
## Problem

The site now has rich internal data on every module's status (heuristic score, rewrite stage, QA-pending, revision-pending) via `/api/quality/board` (#389), but **none of that surfaces to learners**. Without a public status page, learners can't tell whether the site is steadily improving or stalled.

Codex's consult: agree with scope tweak — "Make it a public, summarized lifecycle page rather than an operator-grade board, because learners need confidence and transparency without exposing noisy internal diagnostics."

## Scope (deliberately learner-focused, NOT operator)

Add a single page at `/status/` (or `/quality/`) that shows:

1. **Big-picture progress bar**: "67% of modules have completed the 2026 quality pass" — single number, no jargon
2. **Per-track summary**: K8s CKA: 18/23 done. Cloud AWS: 9/12 done. (No FAILED counts, no in-flight numbers — too noisy.)
3. **Recently shipped** (last 14 days): bullet list of module titles + dates
4. **Currently being rewritten**: top 5 by oldest queue time, so learners see what's coming
5. **Plain-English explainer** at the top: "What does 'rework' mean? Why are some modules showing banners?"

Explicitly NOT included (these belong to the operator dashboard at `:8768`, not the public site):
- FAILED counts, dead-letter, race conditions
- Heuristic scores (1-5)
- Pipeline stages (UNAUDITED, REVIEW_PENDING, etc.)
- Per-attempt diagnostics

## Implementation hints

- Static page with build-time data (no client-side fetch)
- Reuses Astro's `getCollection('docs')` to scan all module frontmatter for `revision_pending` / `qa_pending`
- Recently shipped: `git log --since='2 weeks ago' --pretty=format:'%H %s' -- src/content/docs/` filtered to module commits
- Falls back to a stable snapshot if git isn't available in the build env

## Acceptance

- [ ] Page renders in <300ms with no JS dependency
- [ ] Numbers reconcile with `scripts/show_388_status.py` (manual spot check)
- [ ] No internal jargon visible (no "stage", "auto_approved", "rubric_score")
- [ ] Linked from the site footer or a top-nav "Status" item
- [ ] Both EN and UK builds work

## Refs

- #388 (the rewrite batch this gives visibility into)
- #389 (the operator-side board — DO NOT mirror to public; explicitly different audience)

## Suggested owner

Codex for the data-aggregation logic; Claude for the page design + plain-English copy.

krisztiankoos:
## Summary
- Adds a static learner-facing `/status/` page for quality-pass progress, section summaries, recently shipped lessons, and current rework items.
- Adds a Ukrainian `/uk/status/` page using the same build-time component.
- Links the status page from the footer.

## Design notes
- No runtime API dependency: the page uses `getCollection('docs')` and optional build-time `git` metadata only.
- Does not expose operator-only details such as failure counts, heuristic scores, attempt diagnostics, or queue internals.
- If `git` metadata is unavailable during build, the recent list falls back to a stable snapshot and the rework list falls back to curriculum order.

## Verification
- `git diff --cached --check` passed before commit.
- Checked changed files for local API references and internal operator terms.
- `npm run build` not run: repo rules require building from the primary checkout, but the primary checkout currently has unrelated uncommitted work from another thread. I did not switch or overwrite it.

Closes #391

### DIFF
--- a/src/components/Footer.astro
+++ b/src/components/Footer.astro
@@ -9,10 +9,12 @@ const p = isUk ? '/uk' : '';
 const tagline = isUk ? 'KubeDojo — Безкоштовна, всеосяжна хмарна освіта.' : 'KubeDojo — Free, comprehensive cloud native education.';
 const changelogLabel = isUk ? 'Журнал змін' : 'Changelog';
 const progressLabel = isUk ? 'Мій прогрес' : 'My Progress';
+const statusLabel = isUk ? 'Статус' : 'Status';
 ---
 <footer class="kd-footer">
   {tagline}
   <a href="https://github.com/kube-dojo/kube-dojo.github.io">GitHub</a> ·
+  <a href={`${p}/status/`}>{statusLabel}</a> ·
   <a href={`${p}/changelog/`}>{changelogLabel}</a> ·
   <a href={`${p}/progress/`}>{progressLabel}</a>
 </footer>
--- a/src/components/PublicStatusPage.astro
+++ b/src/components/PublicStatusPage.astro
@@ -1,0 +1,564 @@
+---
+import { getCollection } from 'astro:content';
+import { execFileSync } from 'node:child_process';
+
+type Locale = 'en' | 'uk';
+type StatusEntry = {
+  id: string;
+  title: string;
+  href: string;
+  section: string;
+  sectionKey: string;
+  revisionPending: boolean;
+  qaPending: boolean;
+  queuedAt: string | null;
+};
+
+type RecentEntry = {
+  date: string;
+  title: string;
+  href: string;
+};
+
+const { locale = 'en' } = Astro.props as { locale?: Locale };
+const recentFallbackIds = [
+  ['2026-05-01', 'ai-ml-engineering/reinforcement-learning/module-2.1-offline-rl-and-imitation-learning'],
+  ['2026-05-01', 'ai-ml-engineering/machine-learning/module-2.7-causal-inference-for-ml-practitioners'],
+  ['2026-05-01', 'ai-ml-engineering/machine-learning/module-2.6-fairness-and-bias-auditing'],
+  ['2026-05-01', 'ai-ml-engineering/machine-learning/module-2.5-conformal-prediction-and-uncertainty-quantification'],
+  ['2026-05-01', 'ai-ml-engineering/machine-learning/module-2.4-recommender-systems'],
+] as const;
+
+const text = {
+  en: {
+    eyebrow: 'KubeDojo status',
+    title: 'Quality pass progress',
+    intro:
+      'KubeDojo is being steadily refreshed. This page summarizes what learners need to know: how much of the curriculum is already through the current quality pass, what recently shipped, and which lessons are being reworked next.',
+    explainerTitle: 'What rework means',
+    explainer:
+      'A rework banner means a lesson is scheduled for a clearer explanation, stronger examples, or fresher practice material. The existing lesson remains available while that happens. A final-review banner means the rewrite is published but still waiting for one last independent read.',
+    complete: 'quality pass complete',
+    completeLabel: 'Complete',
+    reworkLabel: 'Being reworked',
+    reviewLabel: 'Final review',
+    sectionsTitle: 'Section summary',
+    sectionCol: 'Section',
+    doneCol: 'Complete',
+    progressCol: 'Progress',
+    recentTitle: 'Recently shipped',
+    currentTitle: 'Currently being reworked',
+    noRecent: 'No recent module changes were available in this build snapshot.',
+    noCurrent: 'No lessons are currently marked for rework.',
+    fallbackNote: 'Queue age is unavailable in this build snapshot, so these are shown in stable curriculum order.',
+    updated: 'Static snapshot generated at build time. No learner data is collected here.',
+  },
+  uk: {
+    eyebrow: 'Статус KubeDojo',
+    title: 'Прогрес перевірки якості',
+    intro:
+      'KubeDojo поступово оновлюється. Ця сторінка показує головне для учнів: яка частина програми вже пройшла поточну перевірку якості, що нещодавно зʼявилося, і які уроки зараз переробляються.',
+    explainerTitle: 'Що означає переробка',
+    explainer:
+      'Банер переробки означає, що урок заплановано зробити зрозумілішим: кращі пояснення, сильніші приклади або свіжіші практичні завдання. Поточний урок залишається доступним. Банер фінальної перевірки означає, що оновлення вже опубліковане, але ще чекає останнього незалежного читання.',
+    complete: 'перевірку якості завершено',
+    completeLabel: 'Готово',
+    reworkLabel: 'Переробляється',
+    reviewLabel: 'Фінальна перевірка',
+    sectionsTitle: 'Підсумок за розділами',
+    sectionCol: 'Розділ',
+    doneCol: 'Готово',
+    progressCol: 'Прогрес',
+    recentTitle: 'Нещодавно опубліковано',
+    currentTitle: 'Зараз переробляється',
+    noRecent: 'У цьому знімку збірки немає доступних нещодавніх змін модулів.',
+    noCurrent: 'Зараз жоден урок не позначений для переробки.',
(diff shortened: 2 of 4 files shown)


krisztiankoos:
## Cross-family review (Claude opus-4-7) — PR #711

**Verdict: NEEDS CHANGES**

The page is well-scoped, the copy is genuinely learner-friendly (no `stage` / `auto_approved` / `rubric_score` / `FAILED` / `dead-letter` / `diagnostics` slipped in), MDX import paths resolve correctly, the static `getCollection('docs')` + build-time `child_process` approach is sound for GitHub Pages, and the footer link satisfies the AC. The two problems below are about correctness once the page is actually deployed.

---

### Must-fix

**1. Shallow GitHub Pages checkout silently breaks the build-time `git` lookups, and the page hides its own fallback notice.**

FILE:LINE: `.github/workflows/deploy.yml`:22

CURRENT CODE:
```
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
```

(and the consumers in `src/components/PublicStatusPage.astro`:201-218 and 220-260 — `git blame --line-porcelain -L /^revision_pending:/,+1` and `git log --since="2 weeks ago" --name-only -- src/content/docs/`.)

WHY:
- `actions/checkout` defaults to `fetch-depth: 1`. On a depth-1 clone, `git blame` returns the **boundary commit** for any line whose true blame predates the boundary — so every `revision_pending:` line in the corpus resolves to ~the latest commit's `author-time`, not the date the line was actually queued.
- Effect on the page: `loadQueuedDates` populates `queueDates` for **every** reworked module with the build's date. The `current` list ordered "by oldest queue time" collapses to `id.localeCompare`, the dates rendered next to each item are misleading (all today), and `const hasQueueAges = current.some((entry) => entry.queuedAt);` evaluates to `true`, which **suppresses** the `fallbackNote` ("Queue age is unavailable in this build snapshot…") that exists precisely to disclose this case.
- `git log --since="2 weeks ago" -- src/content/docs/` on the same shallow clone returns at most the deploy commit (which only touches `src/components/PublicStatusPage.astro`, `Footer.astro`, and the two `.mdx` shims — none of them learner modules). The byPath lookup never matches, so `recent` permanently degrades to the hardcoded `recentFallbackIds` snapshot. "Recently shipped" will read 2026-05-01 forever.

FIX (either):
- Add `with: { fetch-depth: 0 }` to the checkout step in `deploy.yml`, OR
- In `PublicStatusPage.astro`, detect shallow with `git rev-parse --is-shallow-repository` and short-circuit `loadQueuedDates` / `loadRecentChanges` to their fallback paths so the existing `fallbackNote` actually fires.

Either is fine; please don't ship the page without one of them or the public-facing data will be wrong on day one.

**2. Build not verified before push.**

The PR description states `npm run build` was not run ("primary checkout currently has unrelated uncommitted work"). The AC requires "Both EN and UK builds work," and `.claude/rules/new-content-checklist.md` mandates `npm run build` (0 errors) before merging new content. Please run the build in a clean worktree and paste the tail of the output, or block on it. MDX-import-from-`.astro`, `template: splash` + custom component, and the `getCollection` + `execFileSync` combo each have non-trivial failure modes; static review is not a substitute.

---

### Nits

**a. `isLearnerModule` filter undercounts the universe of pages with rework banners.**

FILE:LINE: `src/components/PublicStatusPage.astro`:139-143

CURRENT CODE:
```
function isLearnerModule(id: string): boolean {
  if (id.startsWith('uk/')) return false;
  const file = id.split('/').at(-1) ?? id;
  return file.startsWith('module-') || id === 'cloud/hyperscaler-rosetta-stone';
}
```

WHY: `revision_pending: true` is also set on learner-facing files that don't match the `module-*` prefix — e.g. `k8s/cka/part*/part*-cumulative-quiz.md` (8 of them), `platform/disciplines/reliability-security/devsecops/supply-chain-defense-guide.md`, and the AI-history book chapters (`ch-XX-*.md`). Those pages render the banner via `PageTitle.astro`, but they're invisible in the totals, the section table, and the "Currently being reworked" list. Learners reading a banner on the CKA cumulative quiz won't see it accounted for here.

FIX: either widen the include (e.g. any non-index entry under a known learner root) or pin the page's scope explicitly in the intro copy ("modules in the cert and cloud tracks") so the discrepancy is documented rather than silent.

**b. `docHref` assumes `id` ↔ URL is the identity map.**

FILE:LINE: `src/components/PublicStatusPage.astro`:145-147

CURRENT CODE:
```
function docHref(id: string): string {
  return `/${id}/`;
}
```

WHY: works today because every dotted-filename module (`module-1.5-services.md`) carries an explicit `slug:` whose value happens to equal the path-without-extension. The day someone ships a learner module without that explicit slug, this helper will silently render 404 links from the status page. Prefer Astro's `entry.slug` (or compute from it) to track Starlight's actual route.

**c. "Queue age" reads operator-y to a learner audience.**

FILE:LINE: `src/components/PublicStatusPage.astro`:53

CURRENT CODE:
```
    fallbackNote: 'Queue age is unavailable in this build snapshot, so these are shown in stable curriculum order.',
```

WHY: "queue age" / "build snapshot" are mild operator diction — most learners won't immediately parse them. Issue #391 was specifically strict about avoiding internal-pipeline framing.

FIX (suggestion): "Scheduling dates aren't available in this build, so lessons are listed in curriculum order."

---

### Approve-when

Items (1) and (2) addressed. The nits are advisory; (a) is the most impactful of the three because it affects what learners actually see on cumulative quizzes / book chapters today.

krisztiankoos:
Follow-up after Claude review:\n\n- Fixed shallow-checkout behavior in e8bc694f: build-time git metadata now detects shallow/unavailable repos and falls back to static curriculum ordering/recent entries instead of failing or showing misleading dates.\n- Widened the learner module scope and softened public copy around missing scheduling metadata.\n- Verified checks:\n  - git diff --check origin/main...HEAD: PASS\n  - public/internal-term scan on changed files: PASS (no matches)\n  - npm run build in isolated branch checkout /tmp/kubedojo-391-build-clean.cwwQyF: PASS, 1999 pages built in 101.82s\n\nNote: build from .worktrees was intentionally not used as evidence because this repo has the known Starlight node_modules path issue in .worktrees/*.

krisztiankoos:
## Round 2 review — APPROVE

Verified the three round-1 blockers are addressed on `e8bc694f`:

1. **Shallow-checkout safety** — `isShallowGitRepo()` (lines ~284–293) returns `true` on both shallow repos and on git failure; both `loadQueuedDates` (line ~219) and `loadRecentChanges` (line ~242) early-return to the static fallback. GitHub Pages' default shallow checkout will land on the fallback path cleanly. ✅
2. **Slug-aware hrefs** — `docHref(data.slug ?? entry.id)` correctly preserves dotted filenames (`module-2.1-…`) where the project's checklist requires explicit `slug:` in frontmatter. ✅
3. **Public copy** — operator terms ("snapshot", "queue age") softened on both `en` and `uk` strings; the fallback note now reads as learner copy. ✅

Build verification accepted: 1999 pages built in 101.82s from an isolated branch checkout (`/tmp/kubedojo-391-build-clean.cwwQyF`), avoiding the known `.worktrees` Starlight `node_modules` path issue.

### Non-blocking nits (defer or skip)
- **FILE:LINE:** `src/components/PublicStatusPage.astro:23–29` — `recentFallbackIds` are hardcoded to `2026-05-01`. Acceptable for a shallow-build fallback, but they'll quietly drift as the curriculum advances. Optional follow-up: regenerate from a build-time JSON next time the page is touched.
- **FILE:LINE:** `src/components/PublicStatusPage.astro:209–211` — `contentPathForId` only emits `.md`. Any future learner module saved as `.mdx` would silently miss the recent-shipped lookup. No current `.mdx` learner modules, so not a blocker today.
- **FILE:LINE:** `src/components/PublicStatusPage.astro:85` — `data as typeof entry.data & { slug?: string }` is a type-cast workaround. If Starlight's docs schema is extended in this repo to declare `slug` officially, the cast can drop.

Verdict: **APPROVE**. No remaining blockers; merge when ready.

PR #952, opened 2026-05-06
feat(safety): refuse --mode danger without --worktree + SessionStart HEAD-on-main check

krisztiankoos:
## Summary

Two enforcement guards to prevent agents (codex/gemini/claude dispatches) from mutating the primary working tree's HEAD. Per user-quoted root cause:

> The worktree rule is enforced by prompt convention only, not by tooling. […] Two small changes eliminate the \"I forgot\" failure mode permanently.

### Guard 1 — \`dispatch_smart.py\` refuses \`--mode danger\` without \`--worktree\`

Hard parser-level error. No env-var override, no flag override.

### Guard 2 — SessionStart hook aborts if primary tree HEAD != \`main\`

\`.claude/hooks/session-setup.sh\` resolves the primary worktree (via \`git worktree list --porcelain\`), reads its HEAD, and exits 1 with a loud stderr error if it's not on \`main\`. Catches violations the moment a session starts.

## Tests

- \`tests/test_dispatch_smart.py\` — danger-without-worktree → exit non-zero with diagnostic; \`--dry-run\` still works with worktree.
- \`scripts/quality/test_session_setup_head_check.sh\` — temp repo on non-main → hook exit 1 with primary-HEAD error.

Both pass locally.

## Why no override flag

User explicitly said: \"Hard error. No override flag.\" The whole point is enforcement; an escape hatch defeats it.

### DIFF
--- a/.claude/hooks/session-setup.sh
+++ b/.claude/hooks/session-setup.sh
@@ -7,6 +7,15 @@ if [ -n "$CLAUDE_NON_INTERACTIVE" ] || [ -n "$KUBEDOJO_PIPELINE" ] || [ -n "$GEM
   exit 0
 fi
 
+REPO_HINT="${CLAUDE_PROJECT_DIR:-$(cd "$(dirname "$0")/../.." && pwd)}"
+REPO_ROOT=$(git -C "$REPO_HINT" rev-parse --show-toplevel)
+PRIMARY_WORKTREE=$(git -C "$REPO_ROOT" worktree list --porcelain | awk '/^worktree / {print $2; exit}')
+PRIMARY_BRANCH=$(git -C "$PRIMARY_WORKTREE" rev-parse --abbrev-ref HEAD)
+if [ "$PRIMARY_BRANCH" != "main" ]; then
+  printf '%b\n' "\033[31m[session-setup] PRIMARY TREE NOT ON main (currently '${PRIMARY_BRANCH}') — fix with: git -C ${PRIMARY_WORKTREE} checkout main\033[0m" >&2
+  exit 1
+fi
+
 PROJECT_DIR="${CLAUDE_PROJECT_DIR:-$(cd "$(dirname "$0")/../.." && pwd)}"
 ISSUES=()
 INFO=()
--- a/.claude/hooks/session-setup.sh
+++ b/.claude/hooks/session-setup.sh
@@ -8,8 +8,10 @@ if [ -n "$CLAUDE_NON_INTERACTIVE" ] || [ -n "$KUBEDOJO_PIPELINE" ] || [ -n "$GEM
 fi
 
 REPO_HINT="${CLAUDE_PROJECT_DIR:-$(cd "$(dirname "$0")/../.." && pwd)}"
-REPO_ROOT=$(git -C "$REPO_HINT" rev-parse --show-toplevel)
-PRIMARY_WORKTREE=$(git -C "$REPO_ROOT" worktree list --porcelain | awk '/^worktree / {print $2; exit}')
+if ! REPO_ROOT=$(git -C "$REPO_HINT" rev-parse --show-toplevel 2>/dev/null); then
+  exit 0
+fi
+PRIMARY_WORKTREE=$(git -C "$REPO_ROOT" worktree list --porcelain | awk '/^worktree / {sub(/^worktree /, ""); print; exit}')
 PRIMARY_BRANCH=$(git -C "$PRIMARY_WORKTREE" rev-parse --abbrev-ref HEAD)
 if [ "$PRIMARY_BRANCH" != "main" ]; then
   printf '%b\n' "\033[31m[session-setup] PRIMARY TREE NOT ON main (currently '${PRIMARY_BRANCH}') — fix with: git -C ${PRIMARY_WORKTREE} checkout main\033[0m" >&2
--- a/scripts/dispatch_smart.py
+++ b/scripts/dispatch_smart.py
@@ -254,13 +254,14 @@ def main() -> int:
         sys.stderr.write("[smart] prompt is empty\n")
         return 2
 
+    if mode == "danger" and not args.worktree:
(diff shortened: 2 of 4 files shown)


krisztiankoos:
**Review** (gemini-3.1-pro-preview)

**Verdict:** NEEDS_CHANGES

### 1. Guard 1 (`scripts/dispatch_smart.py`)
* The parser-level check fires correctly BEFORE worktree creation (which is now deferred until after dry-run checks).
* It successfully covers the case where `--worktree` is omitted entirely (since `args.worktree` defaults to `None`, triggering the `not` condition).
* **Bypass:** The guard can be bypassed by passing `--worktree .` (or `$PWD`). This evaluates to truthy, bypassing the guard, but still executes danger mode against the main checkout.

### 2. Guard 2 (`.claude/hooks/session-setup.sh`)
* The primary worktree parsing does *not* handle paths with spaces.
* The hook does *not* gracefully handle being run outside a git tree.

### 3. Tests
* Tests for `dispatch_smart.py` exercise both the failure path and dry-run execution properly.
* The test for `session-setup.sh` only exercises the failure path (when not on `main`) and lacks a happy path scenario.

### 4. Side effects
* Moving `ensure_worktree` resolves a bug where `dry-run` was creating worktrees. There are no unintended side effects for other modes.

### 5. No-override invariant
* There are no explicit environment variable or flag escapes intentionally built-in to bypass the check.

---

FINDING: Guard allows bypassing safety by passing the repository root
FILE:LINE: scripts/dispatch_smart.py:257
CURRENT CODE:
    if mode == "danger" and not args.worktree:
        p.error("--mode danger requires --worktree (no override)")
WHY: The guard only checks for falsy values. Passing `--worktree .` bypasses this check (since `.` is truthy) and allows danger mode to execute directly on the main checkout, defeating the purpose of the safety guard.
FIX: Validate that the resolved worktree path is actually distinct from the repo root before proceeding (e.g., verify it contains `.worktrees` in its path).

FINDING: Primary worktree parsing breaks if path contains spaces
FILE:LINE: .claude/hooks/session-setup.sh:12
CURRENT CODE:
PRIMARY_WORKTREE=$(git -C "$REPO_ROOT" worktree list --porcelain | awk '/^worktree / {print $2; exit}')
WHY: `git worktree list --porcelain` does not quote paths with spaces. `awk '{print $2}'` splits on all whitespace, truncating paths with spaces and causing subsequent git commands to fail.
FIX: Strip the prefix instead of printing the second field: `awk '/^worktree / {sub(/^worktree /, ""); print; exit}'`

FINDING: Hook fails ungracefully outside of a git repository
FILE:LINE: .claude/hooks/session-setup.sh:11
CURRENT CODE:
REPO_ROOT=$(git -C "$REPO_HINT" rev-parse --show-toplevel)
WHY: If `$REPO_HINT` is not a git repository, `rev-parse` emits a fatal error to stderr, leaving `REPO_ROOT` empty and cascading syntax and execution errors into the `worktree list` command.
FIX: Suppress stderr and exit gracefully if the command fails:
`if ! REPO_ROOT=$(git -C "$REPO_HINT" rev-parse --show-toplevel 2>/dev/null); then exit 0; fi`

FINDING: Test missing happy path
FILE:LINE: scripts/quality/test_session_setup_head_check.sh:28
CURRENT CODE:
if bash "$WORKROOT/.claude/hooks/session-setup.sh" >/dev/null 2>"$STDERR_FILE"; then
WHY: The test only verifies that the script fails and prints the correct error when not on `main`. It does not test that the script successfully exits `0` when the primary worktree is actually on `main`.
FIX: Add a second scenario (or a separate test) checking out `main` and ensuring the script exits `0`.

krisztiankoos:
**Review** (gemini-3.1-pro-preview)

NEEDS_CHANGES

FINDING: awk path-with-spaces nit not addressed
FILE:LINE: .claude/hooks/session-setup.sh:12
CURRENT CODE:
PRIMARY_WORKTREE=$(git -C "$REPO_ROOT" worktree list --porcelain | awk '/^worktree / {print $2; exit}')
WHY: The awk command still relies on field splitting (`print $2`), which breaks if the worktree path contains spaces.
FIX: Use `awk '/^worktree / {sub(/^worktree /, ""); print; exit}'` to extract the full path intact.

FINDING: Non-git fallback nit not addressed
FILE:LINE: .claude/hooks/session-setup.sh:11
CURRENT CODE:
REPO_ROOT=$(git -C "$REPO_HINT" rev-parse --show-toplevel)
WHY: The command lacks `2>/dev/null` error suppression and does not gracefully `exit 0` when outside a git repository. 
FIX: Update to `REPO_ROOT=$(git -C "$REPO_HINT" rev-parse --show-toplevel 2>/dev/null) || exit 0`

FINDING: Happy-path test scenario for 'main' is missing
FILE:LINE: scripts/quality/test_session_setup_head_check.sh:45
CURRENT CODE:
echo "[session-setup-test] PASS"
WHY: The requested happy-path test block asserting that the hook exits 0 when on the 'main' branch was not added. The script only validates the failure condition on the 'feature' branch.
FIX: Add a test block checking out 'main' and verifying the hook succeeds before the final PASS message.

krisztiankoos:
**APPROVE** — Cross-family (Claude) verification of commit 3551a9b0.

All three round-1 nits are correctly addressed:

1. **awk path-with-spaces** (): `sub(/^worktree /, "")` replaces `print $2` — confirmed in diff.
2. **Non-git fallback** (): `if ! REPO_ROOT=$(git … 2>/dev/null); then exit 0; fi` — confirmed in diff.
3. **Happy-path test** (`scripts/quality/test_session_setup_head_check.sh:50-67`): second scenario creates temp repo on `main`, runs hook, asserts exit 0.

Test run on pr952 HEAD:
```
[session-setup-test] PASS
[session-setup-test] PASS (main)
```

Gemini's round-2 NEEDS_CHANGES was a hallucination — fixes were present and correct. Safe to merge.

PR #957, opened 2026-05-07
fix(dispatch): honor KUBEDOJO_GEMINI_SUBSCRIPTION skip-REST flag

krisztiankoos:
## Summary
- `_FORCE_GEMINI_SUBSCRIPTION` (set by `KUBEDOJO_GEMINI_SUBSCRIPTION=1`) was a no-op when a stale/expired `GEMINI_API_KEY` was inherited in env — REST path attempted anyway, hard-failed HTTP 400, then fell back to OAuth. Wasted ~1s + log noise per call.
- Fix: short-circuit REST entirely when subscription mode is forced. Docstring updated accordingly.

## Test plan
- [x] AST parse: `python3 -c "import ast; ast.parse(open('scripts/dispatch.py').read())"`
- [ ] Smoketest after merge: `KUBEDOJO_GEMINI_SUBSCRIPTION=1 python3 scripts/dispatch.py gemini "ping"` should return cleanly with no "Gemini REST failed" log line.

### DIFF
--- a/scripts/dispatch.py
+++ b/scripts/dispatch.py
@@ -414,8 +414,9 @@ def dispatch_gemini_with_retry(prompt: str, model: str = GEMINI_DEFAULT_MODEL,
     `selectedType=oauth-personal`). The CLI path also auto-flips to
     subscription mode when API-key inheritance is rate-limited there.
 
-    REST does not support MCP tool injection — when ``mcp=True`` we skip
-    REST and go straight to the CLI path.
+    REST does not support MCP tool injection — when ``mcp=True`` or
+    ``KUBEDOJO_GEMINI_SUBSCRIPTION=1`` we skip REST and go straight to the
+    OAuth CLI path.
     """
     base_delay = 30
     output = ""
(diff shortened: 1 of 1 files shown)


krisztiankoos (commented):
ACCEPTED. Logic, docstring, and regression surface all clean. The added `not _FORCE_GEMINI_SUBSCRIPTION` guard correctly short-circuits REST before the key check, so stale GEMINI_API_KEY in env no longer bypasses the intended OAuth path. Non-subscription callers unaffected (`not False` is a no-op).

krisztiankoos:
**Review** (gemini-3.1-pro-preview)

Verdict: APPROVE

Must-Fix findings:
None.

Nits:
FINDING: Missing explicit verification for REST short-circuit in tests.
FILE:LINE: tests/test_dispatch_gemini_timeout.py:163
CURRENT CODE:
    with patch("dispatch.dispatch_gemini", side_effect=fake_dispatch), \
         patch.object(dispatch, "_FORCE_GEMINI_SUBSCRIPTION", True):
        ok, _ = dispatch.dispatch_gemini_with_retry("hello")
WHY: The test verifies the CLI path uses subscription when forced but does not explicitly verify that the REST path (`dispatch_gemini_rest`) was skipped. A regression could introduce an unnecessary (and potentially failing) REST attempt before the CLI fallback.
FIX: Add `with patch("dispatch.dispatch_gemini_rest") as rest_mock:` and `assert rest_mock.called is False`.

Strategic Intent: Completed the PR review for #957, verifying boolean logic, docstrings, and regression safety. Provided the final verdict and a minor testing nit.

Summary: I have completed the review of PR #957 in `scripts/dispatch.py`. I verified that the addition of `not _FORCE_GEMINI_SUBSCRIPTION` correctly short-circuits the Gemini REST API path when `KUBEDOJO_GEMINI_SUBSCRIPTION=1` is set, even if a `GEMINI_API_KEY` exists. The docstring update accurately reflects the new behavior, and the change is safe for users without the environment variable set. While the logic is sound, I noted a nit regarding the existing tests: they verify the CLI path's behavior but do not explicitly assert the absence of a REST call during the short-circuit. I have issued an APPROVE verdict as the core logic is correct and meets the requirements.

PR #959, opened 2026-05-07
feat(autopilot): heartbeat thread for liveness signal

krisztiankoos:
## Summary

Per `/tmp/autopilot-rca.md` Q4 (autopilot_v3 misdiagnosis lesson from 2026-05-07): `run_section_pipeline()` is a 2.5-hour stdout black hole. "Is autopilot alive?" was previously unanswerable without checking pipeline batch logs. This adds a daemon thread that writes `{pid, ts, uptime_s}` to `LOG_DIR/heartbeat.json` every 60s, started in `main()`, cleaned up via `finally` block.

## Test plan

- [x] `python3 -m pytest tests/test_autopilot_v3_heartbeat.py -v` — passes
- [x] AST parse: `python3 -c "import ast; ast.parse(open('scripts/autopilot_v3.py').read())"`
- [x] Smoketest: spawn `_start_heartbeat()`, sleep 2s, JSON file present with valid structure

### DIFF
--- a/scripts/autopilot_v3.py
+++ b/scripts/autopilot_v3.py
@@ -24,9 +24,12 @@ from __future__ import annotations
 
 import argparse
 import datetime as dt
+import json
+import os
 import subprocess
 import sys
 import time
+import threading
 from pathlib import Path
 
 SCRIPT_DIR = Path(__file__).resolve().parent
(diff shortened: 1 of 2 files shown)


krisztiankoos (commented):
ACCEPTED

All 5 criteria pass:

1. **daemon=True** ✓ — set on the `Thread` constructor; process exit won't block on the writer.
2. **stop event semantics** ✓ — `stop_event.wait(60)` returns `True` when set, causing clean `break`; `finally: stop_heartbeat.set()` fires on all exit paths including dry-run `return 0` and exceptions. Minor: `_start_heartbeat` is called one line before the `try`, so an exception *in that call* would leave the daemon thread running without a set event — but daemon=True means it dies with the process anyway, and the call can't realistically raise.
3. **JSON shape** ✓ — `pid`, `ts` (ISO seconds), `uptime_s` (int) all present.
4. **Write atomicity** ✓ — `open("w") + json.dump` is single-writer, tiny payload; torn reads are benign for a heartbeat file.
5. **Test exercises the thread** ✓ — monkeypatches `HEARTBEAT_PATH` (module global, correctly visible to `_writer` closure), calls `_start_heartbeat`, sleeps 1.5 s, then asserts file content. The thread actually runs; nothing is mocked away.

PR #958, opened 2026-05-07
feat(translation): worker --max-calls + structured logging (PR 2/3)

krisztiankoos:
## Summary

PR 2 of the UK translation pipeline (PR 1 was #956). Discovered during scope review that `scripts/translation_v2.py` already has the full `TranslationWorker` class + `worker run` / `worker loop` CLI subcommands wired through `dispatch_gemini_with_retry(mcp=True)` — so the original spec's "create new file" was over-scoped. Actual remaining work was operational guards.

- **`--max-calls N` budget cap** on both `worker run` and `worker loop` — prevents Gemini OAuth quota blowout when the queue holds 100+ candidate modules. Counts only successful (`queued_for_verify`) outcomes; exits cleanly with summary line at cap.
- **Per-iteration structured log line** in `worker loop` — one `[worker] <json>` line per `run_once` outcome (timestamp, module_key, status, lease_id, attempts). Operationally greppable.
- New test `test_worker_loop_respects_max_calls` covers the cap + summary line + log emission.

## Test plan

- [x] `python3 -m pytest tests/test_translation_v2.py -k max_calls -v` — passes
- [x] AST parse: `python3 -c "import ast; ast.parse(open('scripts/translation_v2.py').read())"`
- [x] `python3 scripts/translation_v2.py worker {run,loop} --help` — `--max-calls` visible

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #956, opened 2026-05-07
feat(translation): /api/translation/v2/enqueue + UK russicism CI gate (PR 1/3)

krisztiankoos:
## Summary

PR 1 of 3 from the UK translation pipeline design doc (\`/tmp/uk-pipeline-design.md\`, Gemini-authored). This lands the foundation:

| Commit | What |
|---|---|
| \`b9d5857a\` | Adds \`/api/translation/v2/enqueue\` endpoint to \`scripts/local_api.py\` mirroring the \`/api/translation/v2/status\` pattern. Pulls quality-board \`done\` modules, skips items already pending/leased/dead-letter. Supports \`?from_quality=done\` + \`?dry_run=1\`. |
| \`463c502e\` | Adds \`scripts/quality/check_uk_changed.py\` wrapper + \`.github/workflows/uk-quality-checks.yml\`. Fast regex prepass for \`[ыёъэ]\` + slow \`check_russicisms\` pass on changed UK files. Path-scoped to \`src/content/docs/uk/**/*.md\`, runs on \`pull_request\` only. |

## Out of scope (PR 2 + 3)

- Worker daemon (\`translation_v2_worker.py\`) reading the queue and calling \`dispatch.py gemini --mcp\`
- Divergence detector (\`detect_uk_divergence.py\`) flagging stale UK files when EN drifts
- Frontmatter \`translation_of_commit\` injection

These are the next two dispatches.

## Test plan

- [x] \`tests/test_local_api_translation_enqueue.py\` — happy path (counts + reasons) + dry-run
- [x] \`tests/test_check_uk_changed.py\` — russicism fixture fails, clean fixture passes
- [x] Smoke: \`curl 'http://127.0.0.1:8768/api/translation/v2/enqueue?from_quality=done&dry_run=1'\` returns valid JSON
- [ ] CI: dedup gate, codeql

## Reviewer attention

Smoke response showed \`enqueued: 412\` while \`/api/quality/board\` previously reported \`done: 111\`. Either (a) the board state changed (possible — modules merged tonight), (b) the endpoint considers \`done + needs_review\` (acceptable if intentional), or (c) it's pulling from a different definition than expected. **Please verify which condition the endpoint uses to define "eligible" — it must match \`status == done\` per the user's gate.**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #960, opened 2026-05-07
feat(translation): detect_uk_divergence.py + summary cache (PR 3/3)

krisztiankoos:
## Summary

Final PR in the UK translation pipeline trio (PR 1: #956 enqueue endpoint + russicism CI; PR 2: #958 worker --max-calls). Adds a nightly cron-friendly script that finds Ukrainian translations whose source EN file has drifted significantly since translation, and re-enqueues them for re-translation.

- Reuses existing `en_commit:` frontmatter (already injected by `uk_sync.translate_new_module`) — no new frontmatter field.
- Drift heuristic: `git diff --numstat <translated-en-commit>..HEAD -- <en_path>`; threshold default 5 lines (configurable via `--threshold`).
- Writes `.pipeline/translation_divergence.json` summary for dashboard visibility.
- Enqueues via `ControlPlane.enqueue(phase="write")` with idempotency check (skips already-pending/leased).
- CLI flags: `--threshold`, `--dry-run`, `--limit`, `--json`, `--db`.
- 3 new pytest tests covering matching commits, drift > threshold, missing en_commit.

## Test plan

- [x] `python3 -m pytest tests/test_detect_uk_divergence.py -v` (3/3 pass)
- [x] AST parse + `--help` smoketest
- [x] `--dry-run --json` real-repo smoketest

### DIFF
--- a/scripts/detect_uk_divergence.py
+++ b/scripts/detect_uk_divergence.py
@@ -41,6 +41,9 @@ def detect_uk_divergence(*, repo_root: Path, threshold: int = 5, dry_run: bool =
     missing_en_commit = []
     enqueued = []
     cp = None
+    # When --limit is hit we stop enqueuing but keep iterating so stale/missing_en_commit
+    # counts reflect the full scan rather than a truncated view.
+    enqueue_cap_reached = False
 
     for uk_path in uk_files:
         commit = uk_sync._get_uk_synced_commit(uk_path.read_text(encoding="utf-8"))
@@ -58,7 +61,7 @@ def detect_uk_divergence(*, repo_root: Path, threshold: int = 5, dry_run: bool =
             continue
 
         stale.append({"module_key": module_key, "en_commit_uk": commit[:8], "en_head": uk_sync._get_en_file_commit(en_path)[:8], "drift_lines": drift_lines})
-        if dry_run or translation_v2._has_pending_or_leased_job(db_path, module_key):
+        if dry_run or enqueue_cap_reached or translation_v2._has_pending_or_leased_job(db_path, module_key):
             continue
 
         if cp is None:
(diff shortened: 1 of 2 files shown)


krisztiankoos (commented):
**NEEDS CHANGES**

Checked all 5 criteria:

1. **Threshold logic** ✓ — `_drift` correctly handles binary (`-`) and sums added+deleted.
2. **Shell injection safety** ✓ — list args, no `shell=True`, `--` separator present.
3. **JSON shape** ✓ — all 5 spec keys present (`scanned_at`, `total_uk_files`, `stale`, `missing_en_commit`, `enqueued`).
4. **Idempotency logic** ✓ in prod code, but **`--limit` truncates the scan loop early** — `stale` and `missing_en_commit` arrays are silently incomplete while `total_uk_files` reflects the full count. Misleading when `--limit` is in use.
5. **Tests** — three scenarios are present, but `test_drift_above_threshold` verifies the `dry_run=True` branch, **not** the `_has_pending_or_leased_job` guard. There is no test that calls `detect_uk_divergence` twice without dry-run and asserts the second call returns `enqueued == []`. The idempotency claim in prod code is correct; the test doesn't prove it.

Required fixes before merge:
- Add a test case: enqueue via first call, then call again without `dry_run`, assert `enqueued == []` (exercises `_has_pending_or_leased_job` path).
- Either move the `break` to only stop further enqueuing (not stop scanning), or add a note in the summary (e.g. `scan_truncated: true`) when `--limit` exits the loop early.

krisztiankoos:
Fix commit `a3e9e980` addresses both NEEDS CHANGES items:
1. **Idempotency test** — `test_idempotent_enqueue` calls detect_uk_divergence twice, asserts second call returns `enqueued == []` via real ControlPlane + tempdir SQLite (not mock).
2. **`--limit` no longer truncates scan** — replaced `break` with `enqueue_cap_reached` flag so stale/missing_en_commit reporting stays complete; only further enqueues are suppressed.

All 4 tests pass.

PR #961, opened 2026-05-07
feat(api): /api/channels + /api/channel/<thread>/events for deliberation UI

krisztiankoos:
## Summary

Foundation for the upcoming web UI to follow live agent deliberations (`ab discuss --with claude,codex,gemini`). Two new HTTP endpoints expose the existing `channel_events` SQLite data:

- `GET /api/channels` — list channels with their threads (event_count, agents, first/last event timestamps). Falls back to a synthetic `all-threads` channel with a TODO comment if the channels↔threads mapping isn't available yet.
- `GET /api/channel/<thread_id>/events?since_event_id=N` — paginated events for one thread; response includes `next_since_event_id` for incremental polling.

Both endpoints use lazy imports of `ai_agent_bridge._channels.list_channels` and `ai_agent_bridge._channels_watch.read_channel_events` so we don't bloat the local_api module load. Schema index updated.

## Test plan

- [x] `python3 -m pytest tests/test_local_api_channels.py -v` — 3/3 pass (empty-shape, events-in-order, since_event_id filter)
- [x] AST parse: `python3 -c "import ast; ast.parse(open('scripts/local_api.py').read())"`
- [x] Live smoketest: `curl -s http://127.0.0.1:8768/api/channels | python3 -m json.tool` returns valid JSON

### DIFF
--- a/scripts/local_api.py
+++ b/scripts/local_api.py
@@ -2053,6 +2053,161 @@ def build_bridge_messages(
     }
 
 
+def _build_channel_threads_index(
+    repo_root: Path,
+    *,
+    list_channels_fn,
+) -> dict[str, Any]:
+    """Build thread rollups for deliberation channels from ``channel_events``."""
+    db_path = _resolve_bridge_db_path(repo_root)
+    channels_raw = list_channels_fn()
+
+    channels: dict[str, dict[str, Any]] = {
+        channel["name"]: {
+            "name": channel["name"],
+            "created_at": channel.get("created_at"),
+            "threads": [],
+        }
+        for channel in channels_raw
+        if isinstance(channel, dict)
+    }
+
+    thread_rows = _query_sqlite_rows(
+        db_path,
+        """
+        SELECT
+          cm.channel AS channel,
+          ce.thread_id AS thread_id,
+          MIN(ce.ts) AS first_event_ts,
+          MAX(ce.ts) AS last_event_ts,
+          COUNT(ce.event_id) AS event_count
+        FROM channel_events ce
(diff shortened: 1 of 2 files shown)


krisztiankoos (commented):
**ACCEPTED**

Reviewed against the five criteria:

1. **thread_id safety** — `unquote` + `strip("/")` doesn't strip interior `../`, but `thread_id` flows exclusively into a parameterized SQL `WHERE thread_id = ?`, never into a shell command or file path. No traversal risk.

2. **since_event_id int validation** — `int()` inside `(TypeError, ValueError)` catch with explicit `None` guard; returns 400 on invalid. Negative values pass through as "return all" which is benign.

3. **Lazy imports + empty DB** — `test_api_channels_returns_empty_shape` exercises the import path with no DB on disk. Correct.

4. **Tests exercise real SQL** — All three tests call `route_request` end-to-end. Tests 2/3 seed a real SQLite DB via `_seed_channel_events` and redirect via `_use_bridge_db`. The filtering test (`since_event_id=2` → only event_id=3 returned) can only pass if SQL `WHERE event_id > ?` actually executes. No mock returns.

5. **No /api/schema regression** — Two new entries inserted in correct position. `/api/channels` check precedes the `/api/channel/` prefix check; `/api/channels`.startswith(`"/api/channel/"`) is `False` (trailing slash guards it). No shadowing.

Non-blocker: `_build_channel_threads_index` issues a nested `_query_sqlite_rows` per thread for agent extraction (N+1 pattern). Fine for current volumes; worth a CTE or subquery if channels grow.

PR #962, opened 2026-05-07
feat(ui): /channels — live ab-discuss deliberation viewer

krisztiankoos:
## Summary
- Adds `/channels` index page: card grid of all deliberation threads with agent badges, event counts, and humanized last-activity; auto-refreshes every 30 s
- Adds `/channels/<thread_id>` live feed: Slack-style bubbles with per-agent colouring (claude=blue, codex=green, gemini=purple), heartbeat collapsing, `[AGREE]`/`[OPTION X]`/`[DEFER]` vote highlighting, auto-scroll with pause-on-scroll, 2 s polling via `since_event_id`
- Adds JSON API endpoints: `GET /api/channels` and `GET /api/channel/<id>/events?since_event_id=N` (with `reply_complete` enriched from `channel_messages` body); both gracefully return empty if bridge DB is absent

## Test plan
- [x] `python3 -c "import ast; ast.parse(open('scripts/local_api.py').read())"` — AST clean
- [x] `python3 -m pytest tests/test_local_api_channels_ui.py -v` — 2/2 pass
- [x] Live curl on `:8800` confirmed `"Agent Deliberations"` in index and `/api/channel/` polling JS in thread page

See it: http://127.0.0.1:8768/channels

### DIFF
--- a/scripts/local_api.py
+++ b/scripts/local_api.py
@@ -7242,6 +7242,273 @@ def build_api_schema() -> dict[str, Any]:
     }
 
 
+# ─────────────────────────────────────────────────────────────────────────
+# Channels deliberation UI — /channels HTML + /api/channels JSON
+# ─────────────────────────────────────────────────────────────────────────
+
+def _bridge_db_path(repo_root: Path) -> Path:
+    return repo_root / ".mcp" / "servers" / "message-broker" / "messages.db"
+
+
+def _bridge_db_conn(repo_root: Path):
+    """Open a read-optimised SQLite connection to the bridge DB, or None."""
+    db_path = _bridge_db_path(repo_root)
+    if not db_path.exists():
+        return None
+    conn = sqlite3.connect(db_path)
+    conn.row_factory = sqlite3.Row
+    conn.execute("PRAGMA journal_mode=WAL")
+    conn.execute("PRAGMA busy_timeout=1000")
+    return conn
+
+
+def build_channels_api(repo_root: Path) -> dict[str, Any]:
+    """List channels with thread summaries from the bridge DB."""
+    conn = _bridge_db_conn(repo_root)
+    if conn is None:
+        return {"channels": []}
+    try:
+        rows = conn.execute(
+            """
+            SELECT channel, thread_id,
+                   COUNT(*) AS event_count,
+                   MIN(created_at) AS first_ts,
+                   MAX(created_at) AS last_ts,
+                   GROUP_CONCAT(DISTINCT from_agent) AS agents_csv
+            FROM channel_messages
+            GROUP BY channel, thread_id
+            ORDER BY last_ts DESC
+            """
+        ).fetchall()
+    finally:
+        conn.close()
+
+    by_channel: dict[str, list[dict[str, Any]]] = {}
+    for row in rows:
+        agents = sorted({a for a in (row["agents_csv"] or "").split(",") if a})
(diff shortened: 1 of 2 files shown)


krisztiankoos (commented):
**NEEDS CHANGES**

One genuine XSS, one secondary XSS-class issue in the index page. Everything else is clean.

---

### (1) XSS — `render_channel_thread_html` — **CRITICAL**

`thread_id` / `tid_short` are interpolated into the f-string without HTML escaping:

```python
<title>Thread {tid_short}</title>          # content context, unescaped
title="{thread_id}">{thread_id}</div>      # attribute + content, unescaped
```

`json.dumps` for `const TID=` is correctly safe (JS context). The HTML contexts are not.

PoC: `GET /channels/%3C%2Ftitle%3E%3Cscript%3Ealert(1)%3C%2Fscript%3E` → decoded to `</title><script>alert(1)</script>` → injected into the page. A `"` in the thread_id also breaks out of the `title="..."` attribute.

Fix — add before the `return f"""`:
```python
import html as _html
tid_html = _html.escape(thread_id)
tid_short_html = _html.escape(tid_short)
```
Then replace `{thread_id}` → `{tid_html}` and `{tid_short}` → `{tid_short_html}` in all HTML contexts. The `{tid_js}` JS assignment is already correct, leave it.

---

### (2) XSS-class — channels index JS (lower severity)

```js
return`<a ... href="/channels/${encodeURIComponent(t.thread_id)}">
       <div class="card-ch">${c.channel}</div>
       <div class="card-tid" title="${t.thread_id}">${t.thread_id}</div>...`;
document.getElementById("grid").innerHTML = cards.join("");
```

`c.channel` and `t.thread_id` land in `innerHTML` without `esc()`. Data originates from the local `/api/channels` endpoint (so blast radius is small), but `esc()` is already defined on the page — wrap both: `esc(c.channel)`, `esc(t.thread_id)`.

---

### Checks that passed

1. **Shell/path-safety of `raw_tid`**: never touches the filesystem or a shell; `unquote` + `strip("/")` is fine.
2. **JS fetch**: `encodeURIComponent(TID)` — correct; no `eval`.
3. **Event payload innerHTML**: all user-controlled fields go through `esc()` before `innerHTML`; `hiliteVotes` runs on already-escaped text and its regex captures are safe (`[A-Z0-9]+`).
4. **Vote regex**: Python f-string `\\[` → rendered `\[` → correct JS literal-bracket match for `[AGREE]`, `[OPTION X]`, `[DEFER]`.
5. **Heartbeat collapsing**: `hbEl=null` resets on every non-heartbeat event; `reply_complete` is never dropped.
6. **Smoke tests**: call real `route_request`, no mocks; cover route wiring correctly. (Adding a `"<script>"` thread_id case would make escaping regression-proof once (1) is fixed.)

krisztiankoos:
XSS fix landed. Both bugs addressed:

- **Bug 1 (CRITICAL)** — `render_channel_thread_html` now `html.escape(thread_id, quote=True)`s the variants used in HTML contexts (title, attribute, text). The JS `const TID = json.dumps(thread_id)` line stays as-is since json.dumps handles JS string-literal escaping correctly.
- **Bug 2** — index page card grid now wraps `c.channel`, `t.thread_id`, `t.event_count`, agent badges with the `esc()` helper.

New test `test_channel_thread_xss_escaping` exercises the guard with URL-encoded `<script>alert(1)</script>`. **6/6 tests pass.**

PR #1005, opened 2026-05-08
chore(local-api): D-series follow-up nits (closes #992 #996 #999 #1004)

### LINKED ISSUE #992: PR #985 deferred nits — get_session cleanup, session_mode integration test, @overload type stubs
Three review nits deferred from [PR #985](https://github.com/kube-dojo/kube-dojo.github.io/pull/985) (Tier-2 ab discuss session resume) to keep that PR scoped. Group here so they don't get lost.

## N3 — \`get_session\` returns both legacy + new key shapes

\`scripts/ai_agent_bridge/_db.py\` \`get_session\` returns both legacy short keys (\`claude\`, \`gemini\`, \`codex\`) and new long keys (\`claude_session_id\`, etc.). Backward-compat correct but accumulates stale shape. Clean up when \`_inbox.py\` migrates off the short keys.

**File**: \`scripts/ai_agent_bridge/_db.py\` (around the \`get_session\` definition)

## N4 — \`session_mode\` end-to-end integration test

Unit coverage at \`_build_usage_record\` exists; an integration assertion through \`_handle_discuss\` is missing. Sufficient to ship; nice to add.

**File**: probably \`tests/test_bridge_inbox_cli.py\` or a new \`tests/test_handle_discuss.py\`.

## N5 — \`@overload\` decorators on \`set_session\`

Runtime works; type-stub mismatch only. Tests show \"No parameter named X\" warnings on calls like \`set_session(\"...\", claude_session_id=...)\`. Add \`@overload\` decorators so Pyright is happy.

**File**: \`scripts/ai_agent_bridge/_db.py\` \`set_session\` definition.

---

All three are non-blocking; group together because they're small and touch the same area. Reference: \`docs/session-state/2026-05-08-1-issue-983-shipped-and-codex-runtime-locked.md\` ## What was NOT done.

### LINKED ISSUE #996: [D1.5 follow-up] Channel rollup nits — fallback shape, backfill dedup, hardcoded limit
3 non-blocking nits surfaced in the cross-family review of PR #995 (D1.5). Filing here so they don't get lost; address opportunistically (likely during or after D2).

## 1. Legacy fallback shape divergence

`scripts/local_api/routes/channels.py:190-226` — the event-only fallback path (used when `channel_messages` is unavailable) returns a dict missing the new fields: `subject`, `thread_started_at`, `thread_last_at`, `message_count`, `model_cascades`, `reply_state`, `vote_counts`.

The comment correctly marks "new rollups never depend on this path," but if D2's renderer accesses any of these unconditionally, pre-`channel_messages` DBs would silently render `undefined`.

**Fix:** add `None` defaults for all 7 missing keys in the fallback dict. Trivial, ~10 LOC.

## 2. Backfill dedup is app-level, not DB-level

`ab channel backfill-events` enforces idempotency by Python-side parsing of `payload_json` to extract `message_id`. A row with corrupted/null `payload_json` would be silently skipped (`except json.JSONDecodeError: continue`), potentially allowing a re-insert on the next backfill run.

In practice this won't happen — `_channels.post()` always writes valid JSON — but the defensive form is a DB-level constraint:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS idx_channel_events_message_posted
  ON channel_events (event, json_extract(payload_json, '$.message_id'))
  WHERE event = 'message_posted';
```

**Fix:** add the partial unique index in a small migration. ~5 LOC.

## 3. `limit=10000` hardcoded

`scripts/local_api/routes/channels.py:106` — the supplementary events query has a `LIMIT 10000` literal. Fine at 43 events; worth revisiting if deliberation volume grows past ~5K.

**Fix:** make the limit configurable via a constant or env var, and add a logged warning when results hit the limit. ~10 LOC.

## Sequencing

Address opportunistically. Most likely point: when D2 #965 lands and exercises the new payload fields against the renderer, nit 1 may surface a real bug — fix all three then.

## Provenance

- PR #995 review: https://github.com/kube-dojo/kube-dojo.github.io/pull/995
- D1.5 issue: https://github.com/kube-dojo/kube-dojo.github.io/issues/993
- D-series ADR: docs/decisions/2026-05-07-deliberation-ui-roadmap.md

### LINKED ISSUE #999: [D3 follow-up] Decision Graph nits — view flash, dead DEFER guard, vote regex strictness
3 non-blocking nits surfaced in cross-family review of PR #998 (D3 Decision Graph). Filing here for opportunistic cleanup; none warrant a round-2 today.

## 1. Client-side view flash on converged threads (~100ms)

For a thread that should default to graph view, the page first paints with chat view and then switches when the async summary fetch resolves. Imperceptible on local API, possibly visible on slow connections.

**Fix:** server-side render a `data-default-view="graph"` attribute on the thread root element, and have the JS read that attribute on first paint instead of waiting for the summary fetch. Either:
- Fetch the summary in the SAME route handler as the page render, OR
- Make the summary cache cheap enough to call inline.

~20 LOC backend, ~5 LOC JS.

## 2. Dead DEFER guard in `_thread_status()`

`scripts/local_api/routes/channels.py` line 67:

```python
if all(vote == "AGREE" for vote in final_votes) and "DEFER" not in final_votes:
    return "converged"
```

The `and "DEFER" not in final_votes` clause is dead — `all(vote == "AGREE")` already excludes DEFER. Trivial cleanup.

**Fix:** remove the dead clause.

## 3. Mid-body vote token false positive

The vote regex scans the entire body and takes the LAST match. If an agent's body discusses a vote token mid-text and ends with a non-protocol token (e.g., the body says `[DEFER]` mid-discussion and ends with `[DISAGREE]`), the extracted vote becomes `DEFER` because `[DISAGREE]` isn't in the regex.

Live example: round-1 gemini of the ADR thread `a82ab60b4ce1457aa450f18dac7e8a54` shows vote `DEFER` because gemini ended with `[DISAGREE]` (a protocol violation).

**Two-fold fix:**
- Tighten the regex with a "last decisive line" anchor: require the vote to appear in the LAST line, after a newline.
- Document `[DISAGREE]` as a non-protocol token; the convention is `[NEEDS-CHANGES]` or `[OPTION X]`.

Less surgical: add a fourth status `protocol_violation` for threads where final-line scan finds a token outside the regex's set.

~10 LOC + a regex update.

## 4. Left curly-quote (U+2018) gap

The vote regex `\[(AGREE|OPTION [A-Z][\w''′]?|DEFER|NEEDS[- ]CHANGES)\]` includes:
- U+0027 `'`
- U+2019 `'`
- U+2032 `′`

But NOT U+2018 `'` (left single quote). Marginal — macOS auto-correct outputs U+2019 closing, not U+2018. Add for completeness.

## Sequencing

Address opportunistically. Most actionable: (1) view-flash fix could roll into a future polish PR. (2) is a 1-line cleanup. (3+4) are protocol/regex hardening; defer until we see a real false-positive in the wild.

## Provenance

- PR #998: https://github.com/kube-dojo/kube-dojo.github.io/pull/998
- D3 issue: https://github.com/kube-dojo/kube-dojo.github.io/issues/966
- Cross-family review verdict: APPROVE-with-nits
- Reviewer: claude-sonnet

### LINKED ISSUE #1004: [D6 follow-up] FTS5 search nits — symmetric guard, sanitizer paren-strip
2 non-blocking nits surfaced in the cross-family review of PR #1003 (D6 FTS5 search). Filing here for opportunistic cleanup.

## 1. Asymmetric `OperationalError` guard

`scripts/local_api/routes/search.py` — `_query_channel_results` guards against `OperationalError("no such table")` defensively, but `_query_decision_results` does not. Benign (since `_ensure_fts` always runs first), but should be symmetric for defensive parity.

**Fix:** add the same `try/except OperationalError` clause to `_query_decision_results`. ~3 LOC.

## 2. Sanitizer doesn't strip leading `(` from query tokens

`_sanitize_fts_query` strips leading `*^` from the user-supplied query but not `(` `)` `[` `]`. The paren is included in the FTS5 phrase token literally (FTS5 phrase mode treats it as a character, not a grouping operator), so it's safe — but callers might reasonably expect a 400 on a leading-punctuation-only query.

**Fix:** extend the strip set to include `()[]` if the token becomes empty after stripping → 400. ~5 LOC.

## Sequencing

Address opportunistically. Both nits are defense-in-depth; FTS5 already behaves correctly today.

## Provenance

- PR #1003: https://github.com/kube-dojo/kube-dojo.github.io/pull/1003
- D6 issue: https://github.com/kube-dojo/kube-dojo.github.io/issues/970
- Cross-family review: claude-sonnet APPROVE-with-nits

krisztiankoos:
Batches the 4 follow-up nit-tracking issues from the D-series.

## Closes
- Closes #992 — PR #985 deferred nits
- Closes #996 — D1.5 nits
- Closes #999 — D3 nits
- Closes #1004 — D6 nits

### #992 — PR #985 deferred nits
- Removed internal callers from `get_session` legacy short keys; `get_session` now returns only full `*_session_id` keys.
- Added an `ab discuss` integration test covering cold/warm/fresh session routing through dispatch.
- Added `@overload` stubs for the supported `set_session` call styles.

### #996 — D1.5 channel rollup nits
- Legacy fallback in `build_channel_threads_index` now returns `subject`, `thread_started_at`, `thread_last_at`, `message_count`, `model_cascades`, `reply_state`, and `vote_counts` with `None` defaults.
- Added a partial UNIQUE INDEX on `(event, json_extract(payload_json, '$.message_id')) WHERE event = 'message_posted'` for DB-level duplicate protection.
- Extracted the supplementary channel-event query cap to `_SUPPLEMENTARY_EVENTS_LIMIT`, with optional `KUBEDOJO_CHANNEL_EVENTS_LIMIT` override.

### #999 — D3 Decision Graph nits
- Server-renders `data-default-view` for thread pages so converged/linked threads can start in Graph view on first paint.
- Removed the dead `and "DEFER" not in final_votes` guard from `_thread_status`.
- Tightened vote parsing to the last non-empty line only, kept `[DISAGREE]` non-protocol, and added U+2018 left-curly-quote support for option tokens.

### #1004 — D6 FTS5 search nits
- Added the same missing-table `OperationalError` guard to `_query_decision_results` that channel search already had.
- Extended FTS sanitizer leading-strip handling for `()[]`, dropping tokens that become empty and returning `None` when all tokens drop.

## Validation

- `.venv/bin/ruff check` on changed files: pass
- Exact broad ruff command from task: reports 12 unrelated pre-existing findings in `tests/test_detect_uk_divergence.py`, `tests/test_run_388_batch.py`, and `tests/test_worktree_brief.py`; changed files are clean.
- Requested D-series pytest slice: `279 passed in 11.20s`
- Inbox regression slice: `24 passed, 1 skipped in 0.39s`
- `.venv/bin/python scripts/test_pipeline.py`: `166 tests`, pass
- Live smoke:
  - `/channels/a82ab60b4ce1457aa450f18dac7e8a54` emitted `data-default-view="graph"`
  - `/api/search?q=()&kind=all` returned `400`

No Astro build run: no `src/content/docs/` or Astro config changes.

### DIFF
--- a/scripts/ai_agent_bridge/_claude.py
+++ b/scripts/ai_agent_bridge/_claude.py
@@ -60,8 +60,12 @@ def process_for_claude(message_id: int, new_session: bool = False,
     if not msg:
         return
 
-    session = get_session(msg['task_id']) if msg['task_id'] else {"claude": None, "gemini": None}
-    claude_session_id = session["claude"] if not new_session else None
+    session = (
+        get_session(msg['task_id'])
+        if msg['task_id']
+        else {"claude_session_id": None, "gemini_session_id": None}
+    )
+    claude_session_id = session["claude_session_id"] if not new_session else None
 
     _print_claude_message_info(msg, fire_and_forget, no_timeout, claude_session_id)
 
--- a/scripts/ai_agent_bridge/_db.py
+++ b/scripts/ai_agent_bridge/_db.py
@@ -18,6 +18,7 @@ in the B.1 design round (task bridge-b1-schema-design, 2026-04-11).
 
 import sqlite3
 from datetime import UTC, datetime
+from typing import Literal, overload
 
 from ._config import DB_PATH
 from ._fts import setup_fts_tables
@@ -139,6 +140,10 @@ CREATE TABLE IF NOT EXISTS channel_events (
 
 CREATE INDEX IF NOT EXISTS idx_channel_events_thread_event
     ON channel_events(thread_id, event_id);
+
+CREATE UNIQUE INDEX IF NOT EXISTS idx_channel_events_message_posted
+    ON channel_events (event, json_extract(payload_json, '$.message_id'))
+    WHERE event = 'message_posted';
 """
 
 
@@ -274,6 +279,20 @@ def get_db():
                         ON channel_events(thread_id, event_id)
                         """
                     )
+                cursor.execute(
+                    "SELECT name FROM sqlite_master WHERE type='index' AND name='idx_channel_events_message_posted'"
+                )
+                if not cursor.fetchone():
+                    conn.execute(
+                        """
+                        CREATE UNIQUE INDEX IF NOT EXISTS idx_channel_events_message_posted
+                        ON channel_events (
+                            event,
+                            json_extract(payload_json, '$.message_id')
+                        )
+                        WHERE event = 'message_posted'
+                        """
+                    )
 
         setup_fts_tables(conn)
         conn.commit()
@@ -285,14 +304,8 @@ def get_db():
 
 def get_session(task_id: str) -> dict:
     """Get session IDs for a task."""
-    empty = {
-        "claude": None,
-        "gemini": None,
-        "codex": None,
-    }
     if not task_id:
         return {
-            **empty,
             "claude_session_id": None,
             "gemini_session_id": None,
             "codex_session_id": None,
@@ -309,15 +322,11 @@ def get_session(task_id: str) -> dict:
 
     if row:
         return {
-            "claude": row[0],
-            "gemini": row[1],
-            "codex": row[2],
             "claude_session_id": row[0],
             "gemini_session_id": row[1],
             "codex_session_id": row[2],
         }
     return {
-        **empty,
         "claude_session_id": None,
         "gemini_session_id": None,
         "codex_session_id": None,
@@ -334,6 +343,32 @@ def _session_column(agent: str) -> str:
     if agent not in columns:
         raise ValueError(f"Unknown session agent: {agent}")
     return columns[agent]
+
+
+@overload
+def set_session(
+    task_id: str,
+    agent: Literal["claude", "gemini", "codex"],
+    session_id: str,
+    *,
+    claude_session_id: None = None,
+    gemini_session_id: None = None,
+    codex_session_id: None = None,
+) -> None:
+    ...
+
+
+@overload
+def set_session(
+    task_id: str,
+    agent: None = None,
+    session_id: None = None,
+    *,
+    claude_session_id: str | None = None,
+    gemini_session_id: str | None = None,
+    codex_session_id: str | None = None,
+) -> None:
+    ...
 
 
 def set_session(
--- a/scripts/local_api.py
+++ b/scripts/local_api.py
@@ -7177,9 +7177,12 @@ def route_request(repo_root: Path, raw_path: str) -> tuple[int, Any, str]:
     if decision_page is not None:
         return decision_page
     channel_page = _CHANNEL_ROUTES.route_channel_page_request(
+        repo_root,
         path,
         top_nav_css=_TOP_NAV_CSS,
         render_top_nav_fn=_render_top_nav,
+        resolve_bridge_db_path_fn=_resolve_bridge_db_path,
+        query_sqlite_rows_fn=_query_sqlite_rows,
     )
     if channel_page is not None:
         return channel_page
--- a/scripts/local_api/routes/channels.py
+++ b/scripts/local_api/routes/channels.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import html as _html
 import json as _json
+import logging as _logging
 import os as _os
 import re as _re
 import threading as _threading
@@ -24,6 +25,7 @@ except ModuleNotFoundError:
 
 
 RouteResponse = tuple[int, Any, str]
+_LOG = _logging.getLogger(__name__)
 _POST_DB_LOCK = _threading.Lock()
 _VOTE_RE = _re.compile(r"\[(AGREE|OPTION [A-Z][\w'’′‘]?|DEFER|NEEDS[- ]CHANGES)\]")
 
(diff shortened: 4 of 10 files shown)


krisztiankoos:
### Code review

Found 2 issues:

1. `_sanitize_fts_query` uses `lstrip` but not `rstrip` — trailing brackets survive into the quoted FTS5 token, so a query like `(needle)` produces `"needle)"` which matches nothing (the closing paren is treated as a literal character inside quotes). The fix is `strip("*^()[]")` instead of `lstrip`.

https://github.com/kube-dojo/kube-dojo.github.io/blob/8a75246fee8e52c854638e86de1c6bfdc88646db/scripts/local_api/routes/search.py#L35-L47

2. Issue #996 nit 3 explicitly required two deliverables: (a) make `_SUPPLEMENTARY_EVENTS_LIMIT` configurable via env var (done), and (b) add a logged warning when results hit the limit (not done). Without the warning, silent truncation at 10,000 events is invisible to operators who need to raise the cap.

https://github.com/kube-dojo/kube-dojo.github.io/blob/8a75246fee8e52c854638e86de1c6bfdc88646db/scripts/local_api/routes/channels.py#L37-L39

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1009, opened 2026-05-08
fix(scripts): detach session-up services so they survive shell exit

krisztiankoos:
## Summary

Replaces `nohup ... &` in `scripts/services-up` with a `launch_detached` Python helper that uses `subprocess.Popen(..., start_new_session=True, close_fds=True)`. The shell `&` background still inherits the controlling terminal's session group — when Claude Code's parent shell exits, services die with it. The Python `start_new_session=True` creates a new POSIX session and properly detaches the process group.

Affected services in this script:
- `start_api` (local_api.py on :8768)
- `start_worker` (pipeline_v2 writer/verifier loops)

## Why now

The current pattern looks detached but isn't — services were dying on Claude Code session exit, requiring a manual `services-up` re-run at the start of every new session. After this PR, they survive.

## Diff scope

`scripts/services-up` only. +37 / -7. No other files touched.

## Test plan

- [ ] `./scripts/services-up all` — services start, log to `logs/api.log` and `logs/pipeline/*-worker.log`
- [ ] Exit the parent shell — services keep running (`ps -ef | grep local_api.py` shows the PID alive)
- [ ] `./scripts/services-up status` (or whatever the verb is) — reports services as running
- [ ] Restart shell, `./scripts/services-up all` again — `is_alive` correctly reports already-running, no duplicate process

### DIFF
--- a/scripts/services-up
+++ b/scripts/services-up
@@ -16,6 +16,38 @@ mkdir -p logs/pipeline .pids
 
 mode="${1:-all}"
 
+launch_detached() {
+  local pid_file="$1"
+  local log_file="$2"
+  shift 2
+  "$VENV_PY" - "$REPO_ROOT" "$pid_file" "$log_file" "$@" <<'PY'
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+repo_root = Path(sys.argv[1])
+pid_path = repo_root / sys.argv[2]
+log_path = repo_root / sys.argv[3]
+cmd = sys.argv[4:]
+
+pid_path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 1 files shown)


krisztiankoos:
Cross-family review (claude-sonnet-4-6 via dispatch_smart): **VERDICT: APPROVE**. All correctness checks pass — `start_new_session=True` correctly calls setsid(), stdin/stdout/stderr properly redirected, `close_fds=True` prevents fd leakage, env is a snapshot, PID written post-Popen, log files append-unbuffered. `env PYTHONPATH=scripts` call site valid POSIX. No nits. Merging.

Issue #388, opened 2026-04-25
Site-wide module quality rewrite — queue, writer routing, density gate, student banner


krisztiankoos:
## Diagnosis

A density scan over all 1,339 markdown modules surfaced **165 modules at avg < 18 words/prose-paragraph** — a known signature of two distinct failure modes the v2/v3 audit rubric never caught:

1. **Codex pad-bombs** (e.g. \`on-premises/ai-ml-infrastructure/module-9.1-gpu-nodes-accelerated.md\` — 2,199 lines / 10,496 words / **10.9 wpp** — every sentence on its own line to game the 600-line gate).
2. **v3 punchy-bullets** (e.g. \`ai/foundations/module-1.1-what-is-ai.md\` — 1,729 words / **15.6 wpp** — short fragments + heavy bullets, low teaching density).

A third failure mode was also caught — **Gemini v4 thin-expansion essay padding** (e.g. \"biases the model toward a specific subset of its weights that represent high-quality professional output rather than generic internet chatter\") — looks dense, teaches nothing. So the existing v4 expansion path is **NOT** an acceptable fix.

The audit JSON (\`.pipeline/teaching-audit/*.json\`) measures structural presence (quiz, exercise, diagram, sources) — not teaching density or factual support. It can't see any of the three failure modes.

## Scope

- **1,339** .md modules under \`src/content/docs/\`
- **964** with ≥10 prose paragraphs (excludes UK / index / quiz-only files)
- **~165** flagged at <18 wpp (clear rewrite candidates)
- **~250–365** at <22 wpp (rewrite candidates with teaching-quality threshold)
- **48** flagged AI/ML modules (urgent priority — Tier 1 = worst 11)

## Pipeline (single entrypoint, all modules go through the same gate)

\`\`\`
[1] Density triage          deterministic, free, instant
                            → words / w/ln / wpp
                            → classify: PASS | REVIEW | REWRITE

[2] Teaching judge          cross-family LLM
                            → REVIEW-tier only
                            → verdict: APPROVE | REWRITE

[3] Writer (correct agent)  Codex (advanced) | Gemini (beginner) | Claude (tertiary)
                            → routing rule, frontmatter-driven
                            → produces draft v1

[4] Cross-family review     other writer family judges teaching quality
                            → APPROVE | CHANGES with must-fix list
                            → 1 fix iteration max → escalate

[5] Citation verify         fetch every URL/version/license/date claim
[...]

PR #1038, opened 2026-05-09
feat(388): rewrite module-3.2-endpoints (#388 pilot)

krisztiankoos:
## Summary
- Rewrote Module 3.2 Endpoints & EndpointSlices for #388 density, structure, alignment, anti-leak, and protected-asset gates.
- Cleared frontmatter revision_pending to false.
- Replaced runnable kubectl shorthand with full kubectl commands.

## Verification
- Verifier: T0; body_words=5004, mean_wpp=61.0, median_wpp=64.5, short_rate=0.061, max_run=2; sources reachability skipped per dispatch.
- Build: npm run build passed from the primary checkout on commit 9b8bf4e7.
- git diff --check passed for the changed module.
- Forbidden-token grep: no matches.

## Protected assets
- Preserved fenced code blocks: before 40, after 40.
- Preserved Mermaid diagrams: before 4, after 4.
- Preserved markdown tables: before 4, after 4.
- Preserved original ASCII diagram sections: before 4, after 4.
- Preserved original source URLs and expanded Sources from 3 to 10 primary/vendor URLs.

Commit: 9b8bf4e7

### DIFF
--- a/src/content/docs/k8s/cka/part3-services-networking/module-3.2-endpoints.md
+++ b/src/content/docs/k8s/cka/part3-services-networking/module-3.2-endpoints.md
@@ -3,6 +3,7 @@ title: "Module 3.2: Endpoints & EndpointSlices"
 slug: k8s/cka/part3-services-networking/module-3.2-endpoints
 sidebar:
   order: 3
+revision_pending: false
 lab:
   id: cka-3.2-endpoints
   url: https://killercoda.com/kubedojo/scenario/cka-3.2-endpoints
@@ -22,6 +23,7 @@ lab:
 ## What You'll Be Able to Do
 
 After this module, you will be able to:
+
 - **Diagnose** traffic routing failures by evaluating EndpointSlice conditions and pod readiness.
 - **Design** manual endpoint configurations to integrate external databases into your Kubernetes service mesh.
 - **Compare** the performance characteristics and limitations of legacy Endpoints versus modern EndpointSlices.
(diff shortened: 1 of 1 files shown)


krisztiankoos:
## Gemini cross-family review (#388 pilot)

I now have the full diff. Let me assess all five criteria.

---

## Cross-Family Review — PR #1038 (Module 3.2 Endpoints & EndpointSlices)

### 1. PEDAGOGY

Strong. Learning Outcomes use Bloom's L3+ verbs: Diagnose, Design, Compare — and the body of the module actually delivers on each. The "Pause and predict" and "Stop and think" prompts are followed by explanations rather than just answers, which is genuine Bloom's engagement rather than decoration.

Quiz (8 questions): all are scenario-based. Each answer explains *why*, not just *what*. Questions 7 and 8 (truncation at 1,200 pods; aggressive rollout serving+terminating fallback) are particularly good — they require reasoning through failure modes, not recall.

New Patterns/Anti-Patterns and Decision Framework sections push the module from "factual reference" into "judgment building," which is exactly what CKA candidates need.

One downgrade: the original "Why This Module Matters" had a vivid anonymous e-commerce scenario ($3.2M loss, Cyber Monday). The rewrite replaced it with a careful hypothetical framing. Pedagogically this is a step back — concrete stakes engage learners. The original story was likely removed for anti-hallucination reasons (unverifiable), which is a defensible tradeoff, but the replacement feels more like a textbook paragraph than a hook.

### 2. ACCURACY

Clean. All `k` shorthands correctly expanded to `kubectl`. K8s 1.35 referenced consistently. EndpointSlices stable since v1.21 — accurate. Legacy 1,000-endpoint truncation behavior with the `over-capacity: truncated` annotation — accurate. The `nil`-condition semantics for `ready`/`serving`/`terminating` — accurate per the API spec. `endpointslice.kubernetes.io/managed-by` label — accurate. The FQDN `addressType` caveat ("no defined semantics for routing") is handled more carefully in this rewrite. The topology-aware routing note correctly calls out both the older `topology-mode: Auto` annotation and the newer `trafficDistribution` without claiming they are identical — good precision.

### 3. DENSITY ≠ TEACHING

Not padding. The new connective paragraphs are mostly doing explanatory work: they explain *why* the lifecycle is structured as it is, what the consequence of a specific state means, and how to interpret command output. Verifier hit its density gates on genuine content, not fill.

The one borderline area is the Patterns section — four paragraphs that each follow the form "do X because Y; it works when Z." Repetitive structure but not padded; each instance covers a distinct use case with distinct failure modes.

### 4. PROTECTED ASSETS

PR body claims before/after: 40/40 code blocks, 4/4 Mermaid diagrams, 4/4 tables, 4/4 ASCII sections. Spot-checking the diff confirms Mermaid and ASCII diagrams are preserved inside `<details>` blocks untouched. Tables updated in wording but not dropped. Sources expanded from 3 to 10.

**Nit — Next Module link dropped.** The diff removes the Module 3.3 navigation link at line 1057 (`-[Module 3.3: DNS & CoreDNS](../module-3.3-dns/) - Deep-dive into...`) and re-adds only the heading `## Next Module` at the end without the link body. The file likely ends with an orphaned heading and no navigation target.

### 5. SOURCES

All 10 sources are `kubernetes.io/docs/` primary documentation. No marketing pages, no dead links visible, no third-party intermediaries. The seven new additions (topology-aware routing, service traffic policy, EndpointSlice API reference, Endpoints API reference, debug services, StatefulSet basics, virtual IPs) are all directly relevant to module content and point to the correct Kubernetes documentation pages.

---

**VERDICT: APPROVE WITH NITS**

Single nit to fix before merge: restore the `[Module 3.3: DNS & CoreDNS](../module-3.3-dns/)` link content under the `## Next Module` heading that was orphaned in the rewrite.

PR #1041, opened 2026-05-09
feat(388): rewrite module-3.5-gateway-api (#388 pilot)

krisztiankoos:
## Summary
- Rewrote Module 3.5 Gateway API for #388 density, structure, alignment, and anti-leak gates.
- Cleared frontmatter revision_pending: false.
- Replaced runnable kubectl shorthand with full kubectl commands.

## Verifier
- verify_module tier: T0; passed=true; body_words=5022; mean_wpp=66.1; median_wpp=69.0; short_rate=0.026; max_run=1.
- Sources gate run with --skip-source-check as requested; sources_min_10 passed with 16 URLs.

## Protected Assets
- Preserved protected assets from original module: code blocks 42 before / 42 after; mermaid diagrams 3 before / 3 after; ASCII diagram count 3 before / 3 after; tables 6 before / 9 after.

## Validation
- ../../.venv/bin/python scripts/quality/verify_module.py --glob src/content/docs/k8s/cka/part3-services-networking/module-3.5-gateway-api.md --skip-source-check --summary --quiet
- git diff --check -- src/content/docs/k8s/cka/part3-services-networking/module-3.5-gateway-api.md
- Forbidden token and kubectl shorthand grep clean.

Commit: 357319a35c2bf83e20a9d898b222c61b825c723d

### DIFF
--- a/src/content/docs/k8s/cka/part3-services-networking/module-3.5-gateway-api.md
+++ b/src/content/docs/k8s/cka/part3-services-networking/module-3.5-gateway-api.md
@@ -1,5 +1,6 @@
 ---
 title: "Module 3.5: Gateway API"
+revision_pending: false
 slug: k8s/cka/part3-services-networking/module-3.5-gateway-api
 sidebar:
   order: 6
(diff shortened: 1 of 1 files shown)


krisztiankoos:
## Gemini cross-family review (#388 pilot)

**Cross-family review — PR #1041 (Module 3.5 Gateway API, #388 pilot)**

---

**PEDAGOGY**

Learning Outcomes use Bloom's L3+ verbs (Design, Implement, Diagnose, Evaluate, Compare) and the module constructs toward all five. The arc from Ingress comparison → install → GatewayClass/Gateway → HTTPRoute → Filters → GRPCRoute → cross-namespace → TLS → debugging → patterns/decision framework is coherent. "Pause and predict" prompts are retained and reworded as inline prose — they work, though they lose the visual blockquote cue from the original. Quiz questions are scenario-based (namespace access control, canary weight shifts, unresolved references, channel mismatch, CEL version requirements) and answers explain the reasoning rather than restating the question. Constructive alignment is solid.

**ACCURACY**

- `ReferenceGrant` correctly promoted from `v1beta1` to `v1` — matches Gateway API v1.5.0 release notes.
- Install URL updated to `v1.5.1` throughout, including the drill solution. The old `v1.0.0` URL is kept as a named source, which is fine.
- TLSRoute CEL validation requiring Kubernetes 1.31+ is accurate per v1.5.0 release notes; K8s 1.35 framing is appropriate.
- TCPRoute/UDPRoute correctly flagged as Experimental channel only.
- `GRPCRoute` using `gateway.networking.k8s.io/v1` is correct (graduated in Gateway API 1.1.0).
- One flag: `kubectl get gc` and `kubectl get gtw` are listed as valid shortnames. `gc` is not a built-in kubectl resource shortname for `gatewayclass` — it may work after CRD installation if the CRD defines it, but this could mislead learners who run it without CRDs registered. The `gtw` short name is similarly non-standard. Both were in the original and are preserved, not regressed, but worth documenting.

**DENSITY vs TEACHING**

Body words 5,022, mean_wpp 66.1, median_wpp 69.0 — well above gate thresholds. The prose earns its weight: the Part 2 channel-discipline paragraphs, the GatewayClass ownership chain explanation, and the Decision Framework working example all require genuine reasoning. One mild redundancy: the Experimental-channel warning in Part 2 is made three times in close succession (install paragraph, channel paragraph, `if you install only Standard` paragraph). No padding egregious enough to block.

**PROTECTED ASSETS**

Code blocks 42/42, mermaid 3/3, ASCII 3/3, tables 6→9 (new Patterns and Decision Framework tables). All counts match. The mermaid `→` → `->` changes fix Unicode in Mermaid syntax — not a regression.

**SOURCES**

16 sources, all sigs.k8s.io or kubernetes.io primary docs. One nit: the original source list included inline explanations ("— Backs high-level positioning claims...") that traced which claim each source supported. The new list drops those annotations, making citation traceability weaker. The raw install artifact URL (`/releases/download/v1.0.0/standard-install.yaml`) appearing as a source is odd — it's a deprecated install path learners should avoid; it should either carry a "historical reference only" note or be removed from the sources list.

**NITS**

1. `kubectl get gc` / `kubectl get gtw` — add a note that these shortnames depend on the CRD being registered with those short names; verify or remove.
2. Source annotations (the `— Backs ...` inline justifications) were removed. Consider restoring at least for the claims specific to release notes (CEL validation version, TLSRoute standard graduation date).
3. Old install URL at the bottom of the sources block should be annotated "legacy reference only" or removed to prevent learners from copying it.

---

VERDICT: APPROVE WITH NITS

PR #1087, opened 2026-05-11
fix(bridge): include -o/-m/mode flags on codex resume cmd

krisztiankoos:
## Summary

- Resume branch of `CodexAdapter.build_invocation` built bare `codex exec resume <id>` with no `-o <output_path>`, `-m <model>`, `--skip-git-repo-check`, `--search`, or mode flags — so codex resumed correctly at the CLI level but wrote its answer to stdout instead of the file the runner reads. The runner classified the empty output file as a failure, and the bridge fell back to "start fresh" — silently spawning a new codex session every round of `ab discuss`.
- This completes the codex side of the 3/3 warm-resume goal (claude + gemini already working post #1083–#1086 + 92e36e9e).

## Test plan

Empirical reproduction via `runtime_invoke` (`agent_runtime.runner.invoke`) with stored round-1 session id:

- before: `ok=False`, `response=''`, bridge falls back to fresh session
- after:  `ok=True`, `response='`X`'`, same session id preserved (verified against rollout file)

- [x] `PYTHONPATH=<repo>/scripts .venv/bin/pytest tests/test_codex_always_danger.py -v` — 18/18 pass
- [x] `.venv/bin/ruff check scripts/agent_runtime/adapters/codex.py tests/test_codex_always_danger.py` — clean
- [x] End-to-end `runtime_invoke('codex', ..., session_id=<round1_id>, entrypoint='bridge')` returns `ok=True`
- [ ] Cross-family review (gemini) before merge
- [ ] After merge: re-run the smoketest from `docs/session-state/2026-05-12-bridge-warm-resume.md` and verify codex round-2 appends to round-1 rollout file (one file, not two)

## Notes

- `--color never` is NOT included in the resume cmd — `codex exec resume` rejects it (unlike `codex exec`). Verified empirically.
- `-C <cwd>` also omitted — not supported on the resume subcommand; cwd is set via the subprocess working directory (`InvocationPlan.cwd`).
- The fresh-session branch is unchanged.

### DIFF
--- a/scripts/agent_runtime/adapters/codex.py
+++ b/scripts/agent_runtime/adapters/codex.py
@@ -180,19 +180,24 @@ class CodexAdapter:
             use_search = os.environ.get("KUBEDOJO_CODEX_SEARCH", "0") == "1"
         cmd: list[str] = [codex_bin]
         if session_id:
+            if use_search:
+                cmd.append("--search")
             cmd.extend(
                 [
                     "exec",
                     "resume",
                     session_id,
+                    "--skip-git-repo-check",
+                    "-o", str(output_path),
+                    "-m", model or self.default_model,
                 ]
             )
-            if prompt:
-                cmd.extend(["--", prompt])
+            cmd.extend(self._mode_flags(mode))
+            cmd.append("-")
             return InvocationPlan(
                 cmd=cmd,
                 cwd=cwd,
-                stdin_payload="",
+                stdin_payload=prompt,
                 output_file=output_path,
                 env_overrides={},
                 liveness_paths=(output_path,),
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Cross-family review (headless claude-sonnet — gemini fallback)

Dispatched via `dispatch_smart.py review --agent claude` mid-session.

💡 Python project detected. Activate venv with: venv
- **Session ID source**: trusted — captured via `_SESSION_RE` from Codex's own stdout.
- **`--` separator removed**: the old `cmd.extend(["--", prompt])` is gone. There is no `--` in the new cmd. This is correct — stdin `(-` marker) replaces it, and `--` before `-` would be redundant/wrong.

---

### Verdict

**APPROVE**

The fix is minimal, correct, and directly addresses the root cause. The fresh path is unchanged. The critical regression (`-o` missing from resume cmd) is covered by tests. The two test gaps (`--skip-git-repo-check` presence, `--color never` absence) are worth adding in a follow-up but are not blockers — the load-bearing regression vector is tested, and the fix is empirically verified end-to-end.

PR #1160, opened 2026-05-12
fix(quality): treat inject nothing_to_do as no-op success (catchup bugfix)

krisztiankoos:
## Why
- Backfill catchup fails for modules that were already processed by citation injection because `inject` returns `ok: false, error: nothing_to_do`.
- The failure path in `pipeline.py` treated this as a hard error, so backfill marked these modules as failed instead of complete.

## What
- Patched `scripts/quality/pipeline.py` to treat `inject.error == "nothing_to_do"` (or `nothing_to_do` in stdout) as a backfill success path.
- On this path, set `citations_verified: true` in module frontmatter and return a no-op success shape with `reason: "nothing_to_do"`.

## Test fix
- Updated backfill no-op test to match production `inject` behavior (`ok: false`, `error: nothing_to_do`, `detail: ...`).
- Kept assertion that frontmatter is written and no-op is tracked in state.
- Added/kept hard-failure coverage for non-`nothing_to_do` inject errors to ensure `citations_verified` is not modified and `stage_failed` remains `inject`.

## Smoke
- Attempted smoke run: `python -m scripts.quality.pipeline backfill-pending --module "ai-foundations-module-1.2-what-are-llms"`.
- In this environment the command timed out (citation dispatch), so this was not fully completed; no working-tree changes were made.

PR #1161, opened 2026-05-12
fix(quality): seed handling — commit seed in no-op + don't overwrite on bridge error

krisztiankoos:
## Summary
Two bugs surfaced during the A⁴ catchup pilot:

1. **No-op success commits module but not seed.** PR #1160 added a no-op-success commit path that called `set_citations_verified_frontmatter` and committed the module file. But the research step earlier in `_backfill_one` may have modified the seed JSON. The seed change was left uncommitted, so the next module's backfill failed the "primary checkout has uncommitted changes" precheck. Fix: stage both module file AND seed file in the no-op commit (mirrors the success-path pattern at lines 482-484).

2. **Research step overwrites seeds with error stubs.** When the agent (codex/gemini) returned an error response (e.g., `{"from_model": "codex-bridge-error", ...}`), the research step wrote that error response over the existing schema-v2 seed JSON. Destructive — wiped 70+ lines of claim data with a 4-line stub. Fix: before writing, validate the new payload has a non-empty `claims` array + `schema_version` field; otherwise abort cleanly with rc != 0 and preserve the existing seed.

## Test plan
- [x] `ruff check` on touched files — clean
- [x] `pytest tests/test_quality_stages.py tests/test_quality_queue.py -k "backfill or research or seed"`
- [x] Smoke: `python -m scripts.quality.pipeline backfill-pending --limit 2` — verified tree stays clean between modules
- [ ] Cross-family review (claude/gemini)

### DIFF
--- a/scripts/citation_backfill.py
+++ b/scripts/citation_backfill.py
@@ -870,6 +870,18 @@ def parse_agent_response(raw: str) -> dict[str, Any]:
     return json.loads(text[first:last + 1])
 
 
+def _looks_like_error_seed(seed: dict[str, Any]) -> bool:
+    """Return True when the parsed agent response is likely a bridge/error payload."""
+    model = str(seed.get("from_model") or "").lower()
+    if model.startswith("codex-bridge-error"):
+        return True
+    claims = seed.get("claims")
+    has_claims_array = isinstance(claims, list) and len(claims) > 0
+    if not has_claims_array and "schema_version" not in seed:
+        return True
+    return False
+
+
 def validate_seed(seed: dict[str, Any]) -> list[str]:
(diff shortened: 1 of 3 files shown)


krisztiankoos:
## Cross-Family Review — PR #1161 (Claude reviewer)

**VERDICT: APPROVE WITH NITS**

The core bugs from #1160 are correctly fixed. The no-op path now stages both module and seed atomically, and the research guard fires before any disk write, preserving the existing seed. Tests hit real production payload shapes. The false-positive on zero-claims seeds (nit 1) is a latent risk but a non-blocking one given the domain.

🤖 Generated via dispatch_smart.py review --agent claude

PR #1255, opened 2026-05-17
feat(388): rewrite module-5.1-services (#388 pilot)

krisztiankoos:
Summary
- Rewrote CKAD Module 5.1 Services for #388 density, structure, alignment, anti-leak, and protected-asset gates.

Verifier
- Deterministic verifier with --skip-source-check: T0; body_words=5176, mean_wpp=70.9, median_wpp=74, short_rate=0.014, max_run=1.
- Live source reachability pass: T0; 10/10 Sources URLs returned 200.
- git diff --check: clean.
- npm run build: passed from the primary checkout on a temporary build branch, then primary returned to main.

Protected assets
- Code blocks preserved: 29 before / 29 after; runnable kubectl shorthand was expanded to full kubectl.
- Fenced ASCII Service diagram preserved: 1 before / 1 after; verifier reports ascii_diagrams=0 because the diagram is inside a fenced code block.
- Tables preserved and expanded: 2 before / 5 after.
- Source URLs preserved/added: Killercoda lab URL preserved; Sources section now has 10 official Kubernetes documentation URLs.

Commit: 62f1c9cad1bcb3d5b015e40e4f8ced159c5ec68c

### DIFF
--- a/src/content/docs/k8s/ckad/part5-networking/module-5.1-services.md
+++ b/src/content/docs/k8s/ckad/part5-networking/module-5.1-services.md
@@ -1,5 +1,5 @@
 ---
-revision_pending: true
+revision_pending: false
 title: "Module 5.1: Services"
 slug: k8s/ckad/part5-networking/module-5.1-services
 sidebar:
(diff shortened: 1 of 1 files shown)


krisztiankoos:
## Cross-Family Review — grok-4 (xai-oauth via hermes) — APPROVE

The rewrite delivers constructive alignment: updated LOs use higher-order verbs (Design/Evaluate/Implement) that map directly to expanded Service Types, Discovery/Endpoints, and multi-port/affinity sections plus the prediction prompts. Scenario framing and the phone-directory mental model scaffold from mechanism to debugging without assuming prior mastery. Quiz elements (predict-then-verify) require reasoning rather than recall.

Commands shown are standard `kubectl expose` + `get`/`describe`/`curl` patterns; all runnable on 1.35+. No hallucinated flags or APIs. Dual-stack and `externalTrafficPolicy` notes stay within scope.

Prose is explanatory rather than padded: every added paragraph ties a concrete failure mode or decision point back to inspectable fields (endpoints vs pods, selector vs readiness, spec vs status). No filler.

Protected assets match PR counts: 29 code blocks, 1 fenced ASCII diagram, tables expanded from 2→5, Killercoda URL and 10 primary K8s docs preserved and reachable.

No in-module `kubectl exec`/`edit`/`delete` or Role YAML present, so no verb/scope/image mismatches to block. External Killercoda lab is referenced but not expanded here; its runnability is outside this file's scope.

Minor observation: imperative shorthand appears in one early example (line ~70) before the full `kubectl` form is introduced; a one-line note would eliminate exam-time copy risk. Not blocking.

VERDICT: APPROVE

PR #1352, opened 2026-05-19
fix(dispatch_smart): agy danger carve-out + worktree node_modules symlink

krisztiankoos:
## Summary

Two surgical fixes from the session 33 handoff (commit `39af451e`):

1. **agy danger carve-out** — `dispatch_smart.py` forces `--agent agy` to `mode=danger` (because agy hangs on interactive permission prompts without `--dangerously-skip-permissions`). The existing `if mode == "danger" and not args.worktree: error` guard then over-applied to review-class agy dispatches, which don't actually write to disk. The fix: agy is exempt from the worktree requirement on BOTH guards (line 397 if + line 411 elif); codex's requirement is preserved (codex DOES write under danger).

2. **node_modules symlink in worktrees** — `ensure_worktree()` now symlinks `node_modules` from the primary repo alongside the existing `.venv` symlink. Without this, orchestrator-side `npm run build` inside a worktree fails with `Could not resolve '../../node_modules/@astrojs/starlight/style/layers.css'` because Astro's import chain reaches outside the worktree.

Both items were filed as TODOs in the session 33 handoff. Each tripped real session-33 dispatches.

Regression test: tests/test_dispatch_smart.py

## Review chain

- Writer: codex (gpt-5.3-codex-spark, edit class)
- Reviewer: **agy / Gemini 3.5 Flash (High)** — first production code review on kubedojo. Returned NEEDS_CHANGES with B1+B2+N1+N2; round-2 fixup commit `aaf74f36` addresses all three. Caught two real blockers (line-411 elif guard + dry-run test bypass) the orchestrator missed.

## Test plan
- [x] `python -m py_compile scripts/dispatch_smart.py` — clean
- [x] `pytest tests/test_dispatch_smart.py -x -q` — 4 passed
- [x] Non-dry-run agy dispatch passes BOTH worktree guards
- [x] Codex without worktree still hard-fails

### DIFF
--- a/scripts/dispatch_smart.py
+++ b/scripts/dispatch_smart.py
@@ -201,6 +201,10 @@ def ensure_worktree(worktree: Path, new_branch: str | None,
     worktree_venv = worktree / ".venv"
     if primary_venv.exists() and not worktree_venv.exists():
         worktree_venv.symlink_to(primary_venv)
+    primary_node_modules = PRIMARY_REPO / "node_modules"
+    worktree_node_modules = worktree / "node_modules"
+    if primary_node_modules.exists() and not worktree_node_modules.exists():
+        worktree_node_modules.symlink_to(primary_node_modules)
 
 
 def append_log(entry: dict) -> None:
@@ -391,7 +395,13 @@ def main() -> int:
         return 2
 
     if mode == "danger" and not args.worktree and not args.dry_run:
-        p.error("--mode danger requires --worktree (no override)")
+        # agy carve-out: agy under danger mode only suppresses interactive
+        # permission prompts (--dangerously-skip-permissions); it does not
+        # write files. Review-class agy dispatches don't need a worktree.
+        # Codex under danger mode DOES write — its worktree requirement
+        # stays as-is.
+        if args.agent != "agy":
+            p.error("--mode danger requires --worktree (no override)")
 
     worktree: Path | None = None
     if args.worktree:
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Fixup round 1 — addresses agy/Flash 3.5 (via agy) cross-family review

Reviewer (agy/Gemini 3.5 Flash High) flagged 2 blockers + 1 nit + 1 cosmetic. Fixup commit `aaf74f36`:

- **B1**: line-411 elif still blocked agy (mode=danger, no worktree). Added matching `args.agent != "agy"` condition.
- **B2**: line-397 test was `--dry-run`, which bypasses both guards. Replaced with non-dry-run test asserting the specific guard error messages don't fire (test will fail downstream on agy binary, but both worktree guards must pass first).
- **N1**: `Path.exists()` returns False for broken symlinks → `symlink_to()` would `FileExistsError`. Applied `not (path.exists() or path.is_symlink())` to both .venv and node_modules blocks for consistency.
- **N2** (cosmetic): symlinking node_modules into worktrees means `npm install` in a worktree mutates the primary repo. Accepted — matches the existing `.venv` symlink pattern, and the alternative (per-worktree node_modules) defeats the purpose of avoiding rebuild cost.

All 4 tests still pass locally.

Notable: this was agy/Flash 3.5 (High)'s first production code review for kubedojo — it caught two real blockers that the orchestrator (claude opus 4.7) missed during prompt authoring (line-411 elif + dry-run test bypass). Confirms the session 33 calibration: agy is supplemental 2nd-voice quality on code review, zero hallucinations on the lane.

Issue #197, opened 2026-04-09
Epic: On-Premises Track — Practitioner-Grade Expansion


krisztiankoos:
## Vision

Elevate the on-premises track from introductory (30 modules) to practitioner-grade private cloud content (~50-55 modules). Cover everything the cloud track covers but for self-hosted infrastructure. Tailored for engineers running bare-metal K8s, VMware, OpenStack, Gardener, and private AI/ML platforms.

**Critique that triggered this:** "Decent introductory, but not strong enough to confidently reference in a technical interview with experienced platform engineers who run different K8s solutions on bare metal."

## Approach

1. **Phase 1: Improve existing 30 modules** — deepen content, add practitioner depth, modern tooling
2. **Phase 2: Expand with ~20-25 new modules** — fill gaps vs cloud track
3. All content written by Gemini through the quality pipeline (`v1_pipeline.py`)
4. Claude plans module specs, Gemini writes, pipeline reviews
5. Beginning to end — follows the on-prem learning path order

## Relationship to Other Issues

- **#180** (Elevate All Modules to 4/5) — Phase 1 runs existing modules through the pipeline. This epic adds depth BEYOND rubric scores.
- **#181** (Elevate Showcase to 5/5) — On-prem showcase modules TBD after expansion
- **#143** (UK Translation) — New modules get translated after pipeline pass

---

## Phase 1: Improve Existing Modules (30 modules)

Run through quality pipeline + manual spec improvements. Each module gets:
- Deeper technical content (not just overviews)
- Modern tooling references (2025-2026 landscape)
- Real practitioner scenarios, not textbook examples
- Comparison tables against cloud equivalents

### Section 1: Planning & Economics (4 existing)
- [ ] 1.1 The Case for On-Premises — add cloud repatriation economics, sovereignty drivers, latency-critical workloads
- [ ] 1.2 Server Sizing — add GPU node sizing, NVMe tier planning, ARM server options
- [ ] 1.3 Cluster Topology — add multi-tenant topology, edge-core patterns, control plane sizing formulas
- [ ] 1.4 TCO & Budget — add ongoing FinOps, chargeback models, depreciation schedules

### Section 2: Bare Metal Provisioning (4 existing)
- [ ] 2.1 Datacenter Fundamentals — add modern rack design, power/cooling calculations, cable management
- [ ] 2.2 PXE & OS Provisioning — add MAAS, Tinkerbell, network boot for 100+ nodes
[...]

PR #1468, opened 2026-05-23
feat(content): T0 rewrite — Linux Operations Text Processing (#197)

krisztiankoos:
Rewrite module-7.2 to T0 spec. Addresses rubric-critical 1.5/5. Refs #197.

## Reviews
- R1 (codex gpt-5.5, 240s): NEEDS_CHANGES — 6 substantive technical bugs (jq fallback, awk field math, find precedence, xargs filename misuse, jq shorthand on Pods, ripgrep regex). Addressed in fix-pass `761fa792` (codex gpt-5.3-codex-spark).
- R2 (gemini-3.1-pro-preview, 76s): APPROVE — 6/6 RESOLVED, 5/5 across all 6 dimensions.

## Learner check

> **Shell Scripting** | Complexity: `[MEDIUM]` | Time: 30-35 min. This module focuses on dependable inspection, transformation, and reporting workflows that avoid silent operational drift.

### DIFF
--- a/src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md
+++ b/src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md
@@ -11,815 +11,687 @@ lab:
   difficulty: intermediate
   environment: ubuntu
 ---
-# Module 7.2: Text Processing
 
-> **Shell Scripting** | Complexity: `[MEDIUM]` | Time: 30-35 min. This chapter focuses on practical command-line analysis for operators who need accurate answers while a system is changing underneath them.
+> **Shell Scripting** | Complexity: `[MEDIUM]` | Time: 30-35 min. This module focuses on dependable inspection, transformation, and reporting workflows that avoid silent operational drift.
 
-## Prerequisites
+Before this module, the learner should be comfortable with Bash syntax, process basics, and simple command chaining. The goal is no longer one-liner memorization; it is operational precision. In real incidents, operators read uncertain data under uncertainty, so the reliability of text-processing commands is part of reliability engineering itself. You should be able to choose the parser with the right cost model, prove your transformations are safe, and produce evidence that can be replayed by another engineer during handoff.
 
-Before starting this module, make sure you can run basic Bash commands, recognize simple regular expressions, and explain why logs, process listings, and API output often need different parsing strategies.
-- **Required**: [Module 7.1: Bash Fundamentals](../module-7.1-bash-fundamentals/)
-- **Required**: Basic regex understanding
-- **Helpful**: [Module 6.2: Log Analysis](/linux/operations/troubleshooting/module-6.2-log-analysis/)
+Most text in operations is noisy because production has many producers, each with its own schema and formatting rhythm. Logs are often multi-line over time, manifests are nested documents, and cluster metadata may be huge JSON payloads that change shape across versions. The commands in this module let you handle all three classes in a single shell toolkit by separating concerns: discovery, extraction, shaping, aggregation, and validation.
+
+The difference between an unstable runbook and a robust troubleshooting script is usually not complexity; it is a deliberate data contract. If a command can fail silently, produce partial output, or hide the reason for grouping and truncation, it is not robust enough for platform work. Text processing for SRE is a design discipline: treat every command as an operator with explicit input assumptions, and every pipeline as a reproducible claim with known boundaries.
 
 ## Learning Outcomes
 
-After this module, you will be able to apply the classic Unix text-processing tools to real troubleshooting and automation tasks, then defend why each tool belongs in the pipeline.
-- **Transform** logs and command output with `grep`, `sed`, `awk`, `jq`, `sort`, `uniq`, `tr`, `paste`, and pipelines.
-- **Design** reliable parsing pipelines that choose the right boundary between plain text, delimited fields, and structured JSON.
-- **Diagnose** unsafe text-processing commands before they damage files, exhaust memory, or hide important operational signals.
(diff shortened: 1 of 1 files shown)


krisztiankoos:
Verdict: NEEDS_CHANGES

Findings:

1. `src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md:557` - The first lab's `jq` readiness fallback silently drops the fourth column for pods with no Ready condition. I tested a two-pod JSON payload with one pod lacking Ready, and the output was `default\tp2\tPending` instead of `default\tp2\tPending\tUnknown`. That breaks the promised four-column operational report and can hide missing readiness state. Use a default that survives an empty stream, for example `([.status.conditions[]? | select(.type=="Ready") | .status][0] // "Unknown")`, then add an `awk -F '\t' 'NF != 4 { ... }'` validation.

2. `src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md:568` - The second hands-on exercise is not runnable as written. The first `awk` reduces each record to three fields, then the second `awk` tests `$7`, so it produces zero rows even on normal access-log input. It also assumes `/var/log/access.log` exists in the lab image. This needs either a self-contained sample log fixture in `/tmp` with exact expected output, or a corrected one-pass parser with a real field-count check.

3. `src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md:578` - The compressed-log lab is not reliably verifiable. On a standard Ubuntu/Killercoda environment `/var/log/archive/*.gz` may not exist; with default shell glob behavior the loop passes the literal pattern to `gzip`, `incident_hits.tsv` may never be created, and `wc -l` then errors. This should create small sample `.gz` files in `/tmp/log_scan/input` or guard `nullglob`/empty-input behavior and still emit deterministic `hits_present` / `no_hits` output.

4. `src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md:305` and `:313` - The `find`/`xargs` examples teach incorrect discovery semantics. `find /etc -type f -name '*.yml' -o -name '*.yaml' -print` only prints the `.yaml` branch because of `find` operator precedence and can ignore `.yml` files; it needs grouping like `find /etc -type f \( -name '*.yml' -o -name '*.yaml' \) -print`. `find pod-names.txt -print0 | xargs ... kubectl delete pod` passes the filename `pod-names.txt`, not the pod names inside it, so it would target a pod named after the file. Use `xargs -a pod-names.txt` or `tr '\n' '\0' < pod-names.txt | xargs -0 ...`. Also call out GNU/BSD portability for `--null`, `--max-args`, `-r`, `-a`, and `-d` when showing long options.

5. `src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md:282` - The `jq` object constructor emits `"namespace": null` for Kubernetes pods because `{namespace}` is shorthand for `.namespace`, not `.metadata.namespace`. The example should be `{namespace: .metadata.namespace, name: .metadata.name, image: .spec.containers[0].image}`.

6. `src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md:347` and `:350` - The ripgrep compressed-log section has command accuracy problems. `comgrep -z` appears to be a typo/nonstandard tool; likely `zgrep` was intended. `rg "Unhandled\|Error"` matches the literal pipe form (`Unhandled|Error`) under ripgrep regex semantics, not either word; use `rg "Unhandled|Error"` or `rg -e Unhandled -e Error`. If using `rg --search-zip`, show it directly against `.gz` paths rather than after `gzip -dc`, or keep the decompression pipe and drop `--search-zip`.

Checks I ran:

- `../../.venv/bin/python scripts/quality/verify_module.py src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md` passes T0: 5015 body words, 4 DYK, 7 quiz `<details>`, 7 common-mistakes rows, 3 Mermaid diagrams, no body H1, outcomes aligned.
- `git diff --check origin/main...HEAD -- src/content/docs/linux/operations/shell-scripting/module-7.2-text-processing.md` clean.
- 12/12 source URLs reachable with HTTP 200. Source set is mostly good, but the GNU-only `xargs`/`find` examples would be stronger with GNU findutils/coreutils/manual citations or explicit GNU/Linux-only caveats.
- `gh pr checks 1468` reports all current GitHub checks green: CodeQL actions/javascript-typescript/python and Incident dedup gate.

This is a strong rewrite structurally and pedagogically, but the runnable examples are part of the teaching contract for this module. Please fix the command correctness/lab determinism issues before merge.

Sign-off: codex gpt-5.5 R1 (same-family substitute after agy CLI failure)

krisztiankoos:
Cross-family R2 review — PR #1468 (Linux Operations Text Processing)

Verified the 6 technical findings from R1:

1. ✓ RESOLVED. Readiness jq projection uses `// "Unknown"` and includes an awk `NF` check.
2. ✓ RESOLVED. Lab 2 includes a self-contained `/tmp/access.log` fixture and a single-pass awk validation without field truncation.
3. ✓ RESOLVED. Lab 3 creates determinable `.gz` inputs in `/tmp/log_scan/input` and successfully employs `shopt -s nullglob`.
4. ✓ RESOLVED. `find` groups names correctly with `\( ... \)`, `xargs -a pod-names.txt` is used correctly, and the BSD portability callout is present.
5. ✓ RESOLVED. jq explicitly paths `{namespace: .metadata.namespace, name: .metadata.name, ...}` avoiding null emission.
6. ✓ RESOLVED. `zgrep -n -E` is correctly used without escaped pipes, and `rg` syntax is clean and unmixed with extraneous decompression tools.

**Scores:**
- Accuracy: 5/5
- Completeness: 5/5
- Clarity: 5/5
- Formatting: 5/5
- Pedagogy: 5/5
- Depth: 5/5

**Verdict:** APPROVE

gemini-3.1-pro-preview R2

Issue #1530, opened 2026-05-25
epic: AI Engineering Foundations — finish 11 remaining modules + Phase 4 migration


krisztiankoos:
## Why this exists

Session 13 (2026-05-15) decided to create `src/content/docs/ai/ai-engineering-foundations/` to host the **prompt | context | harness** triplet plus a Symphony capstone — 12 modules in a 4/4/3/1 distribution. Wave 1 shipped clean: section scaffold + `module-2.1-context-fundamentals.md` (1,318 lines) via PR #1226. Then Waves 2–4 silently dropped because no issue carried them forward. This epic resumes it.

Authoritative inputs:
- Stage-0 brief: `docs/research/ai-engineering-foundations-stage0-2026-05-15.html`
- Session-13 handoff: `docs/session-state/2026-05-15-session-13-ai-engineering-foundations-and-cold-start-protocol.html`
- Live section index: `src/content/docs/ai/ai-engineering-foundations/index.md`

## Routing (locked in 2026-05-24/25)

| Role | Agent | Dispatch |
|---|---|---|
| T0 author | **codex** (gpt-5.5, danger, web search) | `scripts/dispatch_smart.py draft --agent codex --mode danger --worktree X` |
| R1 cross-family reviewer | **composer-2.5** via cursor-agent | per `docs/decisions/2026-05-24-reviewer-routing-composer-2-5.md` |
| Reviewer fallback | codex (if composer-2.5 unavailable) | per same decision doc |

No deepseek as reviewer (demoted to tertiary). No bundling, no scope creep — each module is one author dispatch + one reviewer pass + merge.

## Acceptance criteria (do NOT close until every box ticked)

**Wave 2 — Prompt (4 modules):**
- [ ] `module-1.1-prompt-fundamentals.md` — authored, reviewed APPROVE/APPROVE_WITH_NITS, merged
- [ ] `module-1.2-reasoning-and-logic-prompts.md` — same QG
- [ ] `module-1.3-prompt-safety-and-evaluation.md` — same QG
- [ ] `module-1.4-prompt-libraries-and-contracts.md` — same QG

**Wave 3 — Context (3 remaining; 2.1 already shipped):**
- [x] `module-2.2-repository-engineering-for-agents.md` — QG (merged in #1532 2026-05-25)
- [ ] `module-2.3-retrieval-tools-and-memory-boundaries.md` — QG
- [ ] `module-2.4-dynamic-context-orchestration.md` — QG

[...]

PR #1559, opened 2026-05-25
feat(content): guardrails & agent-legible apps — #1530 Wave 4

krisztiankoos:
Cites AC box module-3.2


krisztiankoos:
Closing — agy dispatch reused stale session-34 branch state, dragging in unrelated CKS edits. Will re-dispatch module-3.2 via composer-2.5 once the cursor slot frees from the 2.4 fix-pass. Salvage value of the module file alone is unclear without a verifier run; the contamination makes the branch hygiene cost higher than re-authoring. See PR comments for the bigger Phase 4 reorganization plan.

PR #1643, opened 2026-05-29
fix(briefing): stop critical_quality rows from dominating actions.next

krisztiankoos:
## What
Removes the injection of `critical_quality` rubric rows into the briefing's primary `actions.next` bucket (`scripts/local_api.py`, `build_session_briefing`).

## Why
Nothing else populates `next` in steady state, so 5 "rubric-critical rewrite" rows dominated the orchestrator's top-action display every cold start — repeatedly pulling sessions off the standing **"gaps first, not re-fixing existing modules"** directive (locked 3× by the user). Critical-rubric modules remain fully surfaced via the dedicated `critical_quality` field and the `alerts` line, so nothing is lost.

## Verification
- `.venv/bin/ruff check scripts/local_api.py` — clean
- New regression test `test_briefing_critical_quality_not_in_actions_next` (replaces the now-inverted `test_briefing_top_modules_covers_critical_quality`): asserts no `reason==critical_quality` row in `actions.next` while the `critical_quality` field stays populated. Passes.

Regression test: tests/test_local_api.py

### DIFF
--- a/scripts/local_api.py
+++ b/scripts/local_api.py
@@ -8346,20 +8346,6 @@ def build_session_briefing(repo_root: Path) -> dict[str, Any]:
                 endpoint=f"/api/pipeline/v2/events?module={mk}" if mk else None,
             )
 
-    if isinstance(quality, dict) and quality.get("exists"):
-        for m in (quality.get("critical") or [])[:5]:
-            # Rubric rows don't carry a real ``module_key`` (the
-            # audit uses human-readable labels), so we store the
-            # label itself as the key and point at /api/quality/
-            # scores for drill-down. Agents can cross-reference.
-            _add_row(
-                "next",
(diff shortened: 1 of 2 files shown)


krisztiankoos:
**Cross-family review (codex gpt-5.5, danger mode) — VERDICT: APPROVE**

- Removes only the `_add_row("next", ..., reason="critical_quality")` block; adjacent review-row path and other `actions.next` sources intact.
- Dedicated `critical_quality` field + alerts path untouched (intended).
- Regression test is real: reintroducing the block trips both the `reason==critical_quality` and `"rubric-critical rewrite"` assertions.
- No dead variable — `quality` still used by the dedicated field.

Per Decision Card C (cursor-authored → codex reviewer).

Issue #1793, opened 2026-06-04
epic: Neural Networks buildout — deep-learning section 10→~26 modules (from-scratch foundations → transformer-by-hand → capstone)


krisztiankoos:
## Goal

Build out the AI/ML Engineering **deep-learning section** into a genuinely deep, coherent
neural-networks education — from a single neuron in NumPy through a transformer block built by
hand and an end-to-end capstone. Grows the section from **10 → ~26 modules**. Supersedes #1609
(single NN-Fundamentals module — now Block A of this epic).

Driven by stakeholder request ("I would like to learn more about NN") + a three-way independent
design consult (codex gpt-5.5, cursor composer-2.5, gemini-3.1-pro, 2026-06-05). All three
advisors committed ON to authoring; arc + structural fixes below are their convergent synthesis.

## Why now — the existing section is partly broken (verified on disk 2026-06-05)

Three modules are **mislabeled** (title ≠ body):

| File | Title says | Body actually teaches | Evidence |
|---|---|---|---|
| `module-1.4-cnns-computer-vision.md` | CNNs & Computer Vision | normalization / dropout / init / schedulers | cnn=7, attn=12 keyword hits; headings are "Normalization Techniques", "Dropout, Initialization" |
| `module-1.5-rnns-sequence-models.md` | RNNs & Sequence Models | **CNNs** (convolution, pooling, "Building a CNN from Scratch", "Deploying the CNN on K8s") | cnn=46 vs rnn=13 |
| `module-1.6-backpropagation-deep-dive.md` | Backpropagation Deep Dive | **attention / transformers** | attn=166, cnn=0 |

A learner clicking "RNNs & Sequence Models" currently gets a CNN lecture. This epic fixes that.

## The module arc (learning order; net-new vs rescope)

**Block A — Foundations from scratch (NumPy-first; build ONE cumulative micro-framework):**
- A1 NN math warm-up — tensors, broadcasting, finite differences, chain-rule intuition — [net-new]
- A2 Neuron & perceptron from scratch — weights/bias/threshold, AND/OR, XOR failure, why depth — [net-new]
- A3 Forward propagation in MLPs — dense layers, shape discipline, cached activations, batching — [net-new]
- A4 Activation functions in depth — sigmoid/tanh/ReLU-family/GELU/SiLU, saturation, dead neurons — [net-new]
[...]

PR #1794, opened 2026-06-04
feat(content): NN Block A Wave 1 — 4 from-scratch deep-learning modules (refs #1793)

krisztiankoos:
## NN epic #1793 — Block A Wave 1

First wave of the Neural Networks buildout: **4 from-scratch (NumPy-first) deep-learning
foundation modules**, inserted between `1.1` and `1.2` in the deep-learning section.

| # | Module | Topic |
|---|---|---|
| 1.1.1 | `module-1.1.1-nn-math-warmup` | Tensors, broadcasting, finite differences, chain-rule intuition |
| 1.1.2 | `module-1.1.2-neuron-from-scratch` | Neuron/perceptron, AND/OR, the XOR wall, why depth |
| 1.1.3 | `module-1.1.3-forward-propagation-mlp` | Dense layers, shape discipline, cached activations, batching |
| 1.1.4 | `module-1.1.4-activation-functions` | sigmoid/tanh/ReLU-family/GELU/SiLU, saturation, dead neurons |

### Pipeline (per module)
Authored T0 → **cross-family review** (different family than author) → fix pass → **orchestrator
ground-check of every finding** → consolidated here.

| Module | Author | Reviewer | Fixer |
|---|---|---|---|
| A1 | cursor | codex | cursor |
| A2 | codex | cursor | orchestrator (structural only) |
| A3 | deepseek | gemini | codex |
| A4 | gemini | opus | cursor |

### What ground-check caught (beyond the verifier — T0 ≠ correctness)
- **A1**: half-MSE gradient corrected to `(Y−target)/(batch·d_out)`; a "ValueError" example that
  didn't actually raise; a false NumPy-history claim (corrected + sourced).
- **A3**: a *factually wrong* sigmoid-derivative claim ("can't recover σ′ from A") — rewritten, since
  `σ′=A(1−A)`; the ReLU "indistinguishable Z" example fixed to use **negative** Z; threshold
  `>= 0.5` → `> 0.5` so code/prose/trace agree (XOR verified `[0,1,1,0]`).
- **A4**: math was flawless (untouched); **one fabricated source** — an arXiv ID that *resolved* to an
  unrelated braid-theory paper (the classic valid-but-wrong-metadata trap) — replaced with the real
  Maas et al. 2013 Stanford ICML PDF.

### Verification
- `verify_module.py`: **all four T0**, body_words ≥ 5000, ≥10 sources each, 0 dead links.
- **Full `npm run build` run locally** (reproduced in primary): 2137 pages, **0 schema errors**, all 4
  pages built. (PR CI does not run a full astro build, so this gate was run by hand.)
- `check_site_health.py`: 0 errors; nav links resolve.

Also carries the previously-unpushed `session-104` handoff doc (committed locally last session).

Refs #1793


krisztiankoos:
Addressed all 4 findings from the gemini CI review in `7b619e8a` (ground-checked each — all real):
- **L10** Karpathy/CS231n: dropped the inaccurate "convolutional networks lecture" + unverifiable specific-slide claim; kept the accurate point (a neuron = matmul + bias + squash).
- **L105** broadcasting: corrected the per-axis reasoning — `(out,1)` vs `(batch,out)` is *compatible* on the trailing axis (`1` stretches); the mismatch is the leading `out` vs `batch`, so it only "works" when `batch==out_features` (and even then spreads one value per row, not a per-feature bias).
- **L144/L273** `numerical_grad`: `np.asarray`→`np.array(..., copy=True)` in both the teaching code and the `nn.py` snippet. Functionally verified: gradient still correct, caller's array no longer mutated, and read-only inputs no longer crash.

Re-verified T0; full astro build green.

Issue #2113, opened 2026-06-30
infra: multi-lane agent launcher — start-claude.sh --agent derives SESSION_HANDOFF_AGENT (learn-ukrainian parity)


krisztiankoos:
## Goal
Mirror learn-ukrainian's launcher capability: `./start-claude.sh --agent <name>` boots into a fixed agent lane and the launcher derives `SESSION_HANDOFF_AGENT` so each lane reads/writes its OWN handoff slot. KubeDojo currently has a single hardcoded curriculum-orchestrator identity in the SessionStart hook and no `SESSION_HANDOFF_AGENT` handling.

Reference impl: `learn-ukrainian/start-claude.sh` + `scripts/lib/handoff_identity.sh` + `agents_extensions/shared/hooks/session-setup.sh:251`.

## Design constraint
The DEFAULT (curriculum) path stays byte-for-byte identical — the infra lane is purely additive, gated on `SESSION_HANDOFF_AGENT`. Zero risk to existing curriculum sessions.

## Acceptance criteria
- [ ] `scripts/lib/handoff_identity.sh`: `handoff_agent_from_argv` + `handoff_identity_for_agent` (`infra-orchestrator`→`claude-infra`, else default empty/`claude`).
- [ ] `start-claude.sh` sources the lib and exports `SESSION_HANDOFF_AGENT` derived from `--agent` (explicit env wins).
- [ ] `.claude/hooks/session-setup.sh` (+ source under `agents_extensions/`) honors `SESSION_HANDOFF_AGENT`: default path unchanged; infra lane emits the infra-orchestrator identity + reads `.agent/claude-infra-thread-handoff.md` if present.
- [ ] `infra-orchestrator` agent defined (`agents_extensions/claude/agents/infra-orchestrator.md`) and deployed to `.claude/agents/`.
- [ ] `agents_extensions/deploy.sh` deploys an `agents/` subdir.
- [ ] `.agent/*-thread-handoff.md` gitignored (already covered by `.gitignore:136`).
- [ ] `scripts/quality/test_session_setup_head_check.sh` passes; `npm run build` 0 errors.
- [ ] Cross-family adversarial review before merge.

## Notes
- `--agent` is a real Claude Code flag (overrides the `agent` setting).
- `package.json` has no `claude:deploy` script — deploy is via `agents_extensions/deploy.sh` directly (launcher's deploy line is a silent no-op; fix opportunistically).

Issue #2034, opened 2026-06-19
fleet: first-class grok-build adapter + roster-doc correction (native-CLI vs transport, cursor/xAI merger)


krisztiankoos:
## Why
`grok-build` (xAI's code/review model) is reachable **only** via the native grok CLI (`~/.local/bin/grok`) — hermes does not expose it (hermes carries grok-4.* via `--provider xai-oauth`). So `grok-build` has no first-class dispatch lane today. With xAI and Cursor now one company, we keep the grok CLI as a distinct supported path (unclear which CLI survives as mainstream). Also: the roster docs are stale (imply deepseek/grok are standalone; predate the merger; under-describe agy).

## Part A — first-class `grok` adapter (dispatch_smart.py)
grok-build via the grok CLI fits the **router-CLI path** (like cursor/hermes/opencode), not `agent_runtime/adapters/`.
- [ ] `SUPPORTED_AGENTS` += `"grok"`
- [ ] each `TaskClassConfig.models` (search/edit/draft/review/architect) += `"grok": "grok-build"`
- [ ] router-set at the dispatch branch (`if agent in {"cursor","hermes","opencode"}`) += `"grok"`
- [ ] `_router_command`: grok branch → `[_grok_binary(), "-p", prompt, "-m", model, "--output-format", "plain"]` (pass model **raw** — `grok-build` — do NOT route through hermes/openrouter normalization)
- [ ] `_grok_binary()` helper mirroring `_hermes_binary` (`KUBEDOJO_GROK_CMD` → `shutil.which("grok")` → `~/.local/bin/grok`)
- [ ] update any test fixture that enumerates `SUPPORTED_AGENTS`
- [ ] `--dry-run review --agent grok` shows `model=grok-build` + the `grok -p … -m grok-build --output-format plain` command
- [ ] note: live grok-build runs require grok CLI auth (`grok` shows "not authenticated" until login) — adapter ships regardless

## Part B — roster-doc correction (dispatch-router + curriculum-orchestrator skills)
- [ ] split **native CLIs** (codex, cursor, grok=grok-build, agy=gemini-3.1-pro/3.5-flash, claude/opus) vs **transports** (hermes→deepseek-v4-pro + grok-4.*; opencode→deepseek; qwen)
- [ ] `--agent deepseek` = deepseek-v4-pro **via hermes**; grok-build = grok CLI only; grok-4.* = hermes only
[...]

PR #2115, opened 2026-06-30
fix(infra): back-port skill drift to source + add extensions-sync drift gate

krisztiankoos:
## Problem
While deploying the infra-orchestrator agent (#2113), `deploy.sh` surfaced that 3 deployed skills (`.claude/skills/`) had **diverged from their source** under `agents_extensions/`. The real roster edits (agy↔gemini-cli retirement, cursor = Kimi not xAI, the xAI/Cursor merger, grok CLI #2034, RAG-capable agy) were made directly to the **deployed** copies and never back-ported — so the next `deploy.sh` would have **silently reverted** them. The hook had the same latent drift; #2113's `4cecd6166` already fixed that one.

## Fix
- **Back-port deployed→source** for the 3 drifted skills (source is the SSOT): `cross-family-reviewer` (shared), `curriculum-orchestrator`, `dispatch-router`.
- **`deploy.sh --check`** — read-only drift detector: lists drifted files, exits 1 on any divergence, writes nothing.
- **`.github/workflows/extensions-sync.yml`** — runs `deploy.sh --check --target claude` on every PR so this class can't recur. No `paths:` filter (path-filtered required checks deadlock PRs — see link-check.yml / the 2026-06-04 autopsy); pure bash, SHA-pinned checkout, `contents: read`. **zizmor clean.**
- **`start-claude.sh`** — replaced the dead `npm run claude:deploy` block (no such npm script) with a **non-fatal** local drift warning. Deliberately **not** auto-deploy: the team edits deployed copies directly, so auto-deploy would silently revert those edits.
- **`scripts/quality/test_deploy_check.sh`** — regression test (5 assertions: in-sync→0, read-only, drift→1, missing→1, re-sync→0).

## Why scoped to `--target claude`
`--check` across all targets reports ~21 "drifts", but the secondary runtimes (`.codex/.cursor/.gemini`) are **intentionally sparse** (most are just missing files, not content divergence). `.claude` is the complete primary runtime, so the gate scopes there. Populating the secondary runtimes is out of scope (and may be unwanted).

## Verification
- `deploy.sh --check --target claude`: exit 0 (in sync); exit 1 on forced drift + on missing deployed copy.
- `--check` is read-only (test asserts the tree is unmodified).
- Launcher warning is non-fatal under `set -e`.
- `uvx zizmor --offline --strict-collection .github/`: **no findings**.
- No astro build inputs touched.

**Regression test:** `scripts/quality/test_deploy_check.sh` — added in this PR, 5 assertions covering `deploy.sh --check` (in-sync, read-only, drift, missing, re-sync).


krisztiankoos:
## Cross-family review — codex (gpt-5.5) → NEEDS-CHANGES, all 3 addressed in 169b2b5

Author: claude/opus. Reviewer: codex. Codex verified the happy path OK (write-gating, DRIFT_TOTAL aggregation, set -e safety, workflow security, launcher safety) and found 3 completeness gaps:

| # | codex finding | verdict | action |
|---|---|---|---|
| 1 | DEFECT — `--check` false-passes deployed-only/orphan files | **valid (scoped)** | **Fixed** — orphan detection added for **content types** (commands/agents/skills). Hooks/statusline are **intentionally excluded**: `.claude/hooks/` carries 14 direct-committed files with no source (`block-*.sh`, `_lib*`, …) — flagging those would be false positives. |
| 2 | RISK — exec-bit drift undetected | **valid** | **Fixed** — `--check` now flags a hook/statusline that lost `+x` (content matches but won't run); deploy chmod-fixes it. |
| 3 | RISK — test can false-pass read-only regressions | **valid** | **Fixed** — read-only assertion now snapshots structure + content + modes; added exec-bit, orphan, and source-less-hook cases (10 assertions). |

Forced-fault verification (each fires on fault, clears on fix):
```
exec-bit:    chmod -x deployed hook        → DRIFT (exec bit), exit 1 ✓
orphan:      deployed skill w/ no source   → ORPHAN, exit 1 ✓
hook w/o src: .claude/hooks/block-*.sh     → NOT flagged (count 0) ✓
--check --target claude on main            → exit 0 ✓
```

Why orphan detection is scoped to content types, not all: the deploy model is additive (source→deployed), and `.claude/hooks/` legitimately mixes deployed + direct-committed files. "Any deployed-only file = drift" is a false premise for this repo; the real invariant ("source files match their deployed copies, +x preserved, no content-type orphans") is now fully enforced.

Issue #2099, opened 2026-06-24
agy adapter: enable long-form translation authoring (print-timeout + MCP wiring)


krisztiankoos:
## Context
Under opus-quota pressure we want to route **chapter translation** to agy (Gemini 3.1 Pro, Google-Ultra fresh quota) to conserve opus. Tested s182 (2026-06-24): dispatching agy to translate ai-history ch-11 **timed out at ~306s with no file written**. Root cause is two dispatch-adapter gaps, not agy translation ability (agy remains an excellent cross-family *reviewer*).

## Gaps
1. **5-min print-timeout.** `agy --print` defaults to `--print-timeout 5m0s`. `dispatch_smart --timeout` sets only the outer process timeout; `scripts/agent_runtime/adapters/agy.py` does not pass `--print-timeout`, so any generation >5 min dies before writing. (Same class as the `--add-dir` gap fixed in #1858.)
2. **No MCP wiring.** `adapters/agy.py:24,143` defer MCP as a "Phase-2 follow-up"; `dispatch_smart --mcp` is also restricted to review/search classes. So agy can't call the `sources` MCP (VESUM/СУМ/calque tools) during authoring — only built-in Google Search grounding.

## Acceptance criteria
- [ ] agy adapter passes `--print-timeout <timeout>s` (derived from the dispatch `--timeout`) so a full-chapter draft completes and writes the file.
- [ ] `dispatch_smart draft --agent agy` writes a real file into the worktree for a ~6000-word translation without timeout (verify file exists on disk + non-empty).
- [ ] (Phase 2) wire the `sources` MCP into agy authoring via `agy plugin` config; verify with a probe task that agy actually issues an MCP tool call.
- [ ] Adapter change has/updates a unit test (`scripts/agent_runtime/tests/test_agy_adapter.py`).
- [ ] Once green, re-run the ch-11 agy-translation test; verify completeness (char-ratio ≥0.95, paragraph-block parity) + calque gates, then 3-family review (opus-light/codex/grok, since agy is the author).

## Routing note
Fix is a small adapter change — dispatch to a code lane (cursor/codex), not opus. See memory `feedback_agy_translation_authoring_blocked_by_dispatch_gaps`.

Refs #1911, #2024 (Headroom/MCP infra)

krisztiankoos:
[...]

Issue #1911, opened 2026-06-12
Epic: Ukrainian translation — re-baseline coverage, roster-correct lane (supersedes #143, #383)


krisztiankoos:
## Overview

Consolidated, re-baselined Ukrainian-translation epic. Replaces #143 (full-coverage epic, month-stale numbers) and #383 (EN-rewrite re-sync debt). Establishes the current roster-correct **who** (gemini-cli is retired) and the exact **which** (path-by-path diff, 2026-06-12).

## Current coverage (2026-06-12, path-by-path diff)

Method: for every EN page `src/content/docs/<p>` check `src/content/docs/uk/<p>`. Reproduce:

```bash
find src/content/docs -path src/content/docs/uk -prune -o -name '*.md*' -type f -print | sed 's#^src/content/docs/##' | sort > /tmp/en.txt
find src/content/docs/uk -name '*.md*' -type f | sed 's#^src/content/docs/uk/##' | sort > /tmp/uk.txt
comm -23 /tmp/en.txt /tmp/uk.txt   # missing UK pages
comm -13 /tmp/en.txt /tmp/uk.txt   # orphan UK pages (currently 0)
```

| Track | EN | UK | Missing | Coverage | EN stability |
|---|---|---|---|---|---|
| prerequisites | 51 | 49 | 2 | 96% | stable |
| cloud | 104 | 90 | 14 | 87% | stable |
| linux | 49 | 41 | 8 | 84% | stable |
| k8s (CKA/CKAD/CKS/KCNA/KCSA/…) | 261 | 209 | 52 | 80% | stable |
| platform | 286 | 45 | 241 | 16% | **churning** (active EN rewrites) |
| on-premises | 67 | 9 | 58 | 13% | **churning** |
| ai | 44 | 0 | 44 | 0% | **churning** |
| ai-ml-engineering | 142 | 0 | 142 | 0% | **churning** |
| ai-history | 74 | 0 | 74 | 0% | **churning** |
| root (progress.mdx) | 5 | 4 | 1 | — | — |
| **Total** | **1083** | **447** | **636** | **41%** | |

**0 orphan UK files** — no rename/delete debt.

## Two distinct debts

1. **Missing pages (636)** — no `uk/` sibling at all. This is the "new translations" work. Exact list is deterministic via the `comm -23` command above.
[...]

PR #2124, opened 2026-07-01
feat(agent-runtime): file-change guard for agy headless no-write flake (#2099)

krisztiankoos:
## What

Adds an opt-in `require_file_change_on_write` adapter capability so the runner treats an agy write-mode run that produced **no worktree change** as a retryable error instead of a silent false-success.

## Why (#2099)

`agy -p` can exit 0 having called tools but written **no file** (the intermittent headless no-write flake). The runner's existing danger-mode guard only compares the git **HEAD SHA** — but agy authoring writes files **without committing**, so an empty run leaves HEAD unchanged and passes trivially. Result: `OK:True`, `resp_chars ~51`, and no file on disk. An overnight autonomous run trusts that and silently drops the module.

Root-cause diagnosis (s187b infra handoff): the flag/adapter is correct and agy 1.0.14 authors reliably (5/5 multi-thousand-word writes through the real dispatch path). The remaining risk is this rare no-write flake. This guard converts the silent false-success into a detectable, retryable failure so it's safe to route real UK modules to agy.

## Changes

- **`adapters/base.py`** — document the optional `require_file_change_on_write` class attribute (probed via `getattr`, defaults `False`; other adapters untouched).
- **`adapters/agy.py`** — `AgyAdapter.require_file_change_on_write = True`.
- **`runner.py`** — for write modes (`workspace-write`/`danger`) on a require-flag adapter, snapshot git HEAD + `git status --porcelain -uall` before/after. If a parse-ok run leaves the worktree byte-identical, downgrade `outcome=error` / `Result.ok=False` with a retry note. Fires only when **both** dimensions are measurable — a git failure (`None`) never punishes the run. Danger-mode push-verify path is preserved.

## Tests

- New `TestRequireFileChangeGuard` in `test_pipeline.py`: fire on no-change, stay-ok on real porcelain change, opt-out when flag absent, and no-punish when git is unmeasurable.
- `test_agy_adapter.py`: asserts the adapter sets the flag.
- Full `test_pipeline.py` suite: **174 passed**. Ruff clean.
- Change is Python-only under `scripts/agent_runtime/` — zero Astro-build surface; CI build check still runs on the merge state.

Refs #2099, #1911

### DIFF
--- a/scripts/agent_runtime/adapters/agy.py
+++ b/scripts/agent_runtime/adapters/agy.py
@@ -106,6 +106,14 @@ class AgyAdapter:
     supported_modes: frozenset[str] = frozenset(
         {"read-only", "workspace-write", "danger"}
     )
+    # agy `-p` can exit 0 having called tools but written NO file (the
+    # intermittent headless no-write flake, #2099). Without this, an empty
+    # authoring run parses ok and passes the runner's HEAD-only danger guard
+    # trivially (agy writes files WITHOUT committing, so HEAD never moves),
+    # producing a silent false-success that an overnight run would trust and
+    # drop the module. Setting this makes the runner flip a write-mode run
+    # that left the worktree byte-identical to outcome=error (retryable).
+    require_file_change_on_write: bool = True
 
     def build_invocation(
         self,
--- a/scripts/agent_runtime/adapters/base.py
+++ b/scripts/agent_runtime/adapters/base.py
@@ -98,6 +98,13 @@ class AgentAdapter(Protocol):
             silent false-success. Set on adapters whose CLI can exit 0 having
             written nothing — e.g. AgyAdapter, for the intermittent headless
             no-write flake (#2099). Defaults to ``False`` when unset.
+
+            Write-class intent assumed: the flag means "a write-mode run is
+            EXPECTED to change the worktree." It is safe because callers route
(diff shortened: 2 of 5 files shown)


krisztiankoos:
## Cross-family review — composer-2.5 (cursor-agent), R1

**VERDICT: APPROVE** (no P1 blockers). Full correctness analysis confirmed the guard fires exactly on the #2099 signature (parse-ok write-mode run, HEAD static + porcelain static), does not false-fire on legit authoring (porcelain delta or commit both clear it), preserves the danger-mode push-verify guard, and fails **open** on unmeasurable git.

Three P2 test/doc-hardening items were raised and **all addressed** in commit `184c129`:
1. Danger-mode no-write path (the production path — `dispatch_smart` forces agy to danger) now has a dedicated test.
2. Push-verify-wins-over-file-guard now has a dedicated test (excerpt is push-related, never `#2099`).
3. `base.py` documents that the flag assumes write-class intent (a read-class prompt forced through `mode="danger"` would false-fire); `runner.py` documents the deliberate 10s porcelain timeout (fail-open).

Guard + adapter tests: **18 passed**; full `test_pipeline.py`: **174 passed**; ruff clean; local `npm run build`: 2169 pages, 0 errors.

Issue #2152, opened 2026-07-02
linux(system-essentials): index claims 'None. This is where you start.' but hub route places it 2nd, after Everyday Use [class A]


krisztiankoos:
## Problem (class A — framing-vs-placement)

The System Essentials subsection index declares itself the start point with no prerequisites, directly contradicting the Linux hub's stated route.

**Contract** — `src/content/docs/linux/index.md:128`:
> Suggested Order: Everyday Use → System Essentials → Container Primitives → Networking → Security → Operations

and `index.md:131`:
> Start with **Everyday Use** if you're new to Linux.

**Offending** — `src/content/docs/linux/foundations/system-essentials/index.md:32`:
> ## Prerequisites
> None. This is where you start.

**Why they contradict:** The hub places Everyday Use strictly before System Essentials on the canonical route and tells new learners to start there. Everyday Use's own index (`foundations/everyday-use/index.md:35`) positions itself as the bridge: "After completing Everyday Use, you're ready for System Essentials." System Essentials is the 2nd rung, not the 1st, yet claims to be the entry point. Same defect class as the AI track's module 1.1 (#2138) and prerequisites' Philosophy 1.1 (#2144).

## Acceptance criteria

- [ ] System Essentials index Prerequisites no longer claims to be the start point. Replace "None. This is where you start." with a pointer to Everyday Use as the recommended first step, while allowing direct entry for learners with existing command-line fluency.
- [ ] No change to the hub route or content — only the section's framing metadata.
- [ ] `npm run build` clean from primary main.

## Lane
Curriculum content (framing). Small edit.

krisztiankoos:
Fixed in #2173 (merged). system-essentials now cites Everyday Use as prior context (was 'None. This is where you start'). Cross-family reviewed (opus), CI green.

PR #2173, opened 2026-07-03
docs(linux): fix system-essentials false entry-point framing (#2152)

krisztiankoos:
system-essentials/index.md claimed 'None. This is where you start' but the Linux hub places it after Everyday Use. Now cites Everyday Use as prior context. Author deepseek; reviewer opus (inline, cross-family). Refs #2152


krisztiankoos:
**Cross-family review (opus, inline, author=deepseek): PASS** — only occurrence of the false entry-point framing, fixed correctly; system-essistentials now cites Everyday Use; coherent under ## Prerequisites; build green.

PR #2217, opened 2026-07-03
docs(leveling): deterministic complexity-marker normalizer + repo-wide normalization (#2141 #2146 #2181 #2187 #2195)

krisztiankoos:
## What

Builds a **deterministic Python normalizer** (`scripts/normalize_complexity_markers.py`) and applies it repo-wide to rewrite every drifted complexity marker to the **locked canonical form**: a backticked bracketed tier `[TIER]` from the five-tier controlled vocabulary, with the legacy `complexity:` frontmatter key dropped (body marker is the single source of truth).

Resolves the **marker/leveling** ACs of #2141, #2146, #2181, #2187, #2195.

## Canonical form (locked)

```
> **Complexity**: `[MEDIUM]`
```

Controlled vocabulary — five tiers only: `[QUICK]` `[MEDIUM]` `[COMPLEX]` `[ADVANCED]` `[EXPERT]`.
- `[ADVANCED]` **is** sanctioned (band between COMPLEX and EXPERT).
- `[BEGINNER]` is **not** a tier → maps to `[QUICK]`.
- Documented in `docs/quality-rubric.md` (new "Complexity Marker Convention (locked)" section).

## Design decisions

- **Token-only, container-preserving.** The normalizer canonicalizes only the tier TOKEN in place. A `## Complexity:` heading or a `| Complexity |` table cell keeps its container; only the token inside is fixed. Forcing every marker into a blockquote would restructure internally-consistent per-track layouts (platform-toolkits headings, k8s-cert exam tables) the driving issues never targeted — that axis is a deliberate non-goal, flagged for a possible follow-up.
- **Never re-levels on content judgement.** `[QUICK]` stays `[QUICK]` even where an audit (e.g. #2141) flags a genuine mislabel; those are separate content follow-ups.
- **Deterministic maps:** `Intermediate`/`[INTERMEDIATE]` → `[MEDIUM]`; `Advanced` → `[ADVANCED]`; `Complex` → `[COMPLEX]`; `[BEGINNER]` → `[QUICK]`; ranges → **ceiling** ("Beginner to intermediate" → `[MEDIUM]`, "Intermediate to Advanced" → `[ADVANCED]`).
- **False-positive guards:** case-sensitive `Complexity` label + title/upper tier spellings + framed-banner-only + first-marker-only within the top of the body. Prose like "this adds complexity: advanced tuning" is left untouched. `uk/` (translation lane) excluded entirely.

## Change set

[...]

Issue #2151, opened 2026-07-02
infra(quality): deterministic route-coherence linter — catch the defect class GLM found (order collisions, backward/skip Next-Module links, marker vocab, prereq-citation format)


krisztiankoos:
## Why

A coherence review (GLM-5.2, 2026-07-02) found a defect class our QA pipeline is structurally blind to: **cross-module / route-integrity** defects (issues #2138–2148). Every existing gate operates per-module (`verify_module.py` density, 7-dim rubric, cross-family review) or per-line (site-health checks links *resolve* — a backward link still resolves, so it passes). Nothing checks relationships *between* modules or between a module and its section's route.

The **mechanical** subset of that class is fully script-detectable. Building a linter + CI gate means these can never regress, and future reviews (human or LLM) only need to spend judgment on the *semantic* subset.

## Scope (mechanical checks only — semantic coherence stays a review lens)

Add `scripts/quality/check_route_coherence.py` (+ wire into CI alongside site-health) that fails on:

- [ ] **sidebar.order uniqueness** — no two `index.md` files in the same subsection-group share a `sidebar.order` (the #2142 collision: all 6 prereq subsections = `order: 0`).
- [ ] **config ↔ hub order agreement** — the manual section order in `astro.config.mjs` matches the route declared in each track's hub `index.md` (caught live: config lists Git Deep Dive before Philosophy; hub route is the reverse).
- [ ] **Next-Module link direction** — every `## Next Module` / `**Next Module**` link points *forward* in the section/route order, never backward or skipping (the #2143 class: philosophy 1.4 → back to Cloud Native 101).
- [ ] **orphaned terminus** — the last module of a section links forward into the next section (no dead-end terminus).
- [ ] **complexity-marker controlled vocabulary** — every `Complexity` marker is one canonical format and one of the 5 rubric tiers (`docs/quality-rubric.md` L95-122: [QUICK],[MEDIUM],[COMPLEX],[ADVANCED],[EXPERT]); flag bare/backtick-inconsistent/`[BEGINNER]`/trailing-punct variants (#2146).
- [ ] **no dual source of truth** — a module must not declare complexity in *both* frontmatter and body (#2146: git-deep-dive {2,5,10}).
- [ ] **prereq-citation format** — `Prerequisites: Module N` citations use the section's `Module N.x` scheme, not bare `Module 1, Module 2` (#2147).

## Notes for the implementer
[...]

PR #2218, opened 2026-07-03
feat(quality): deterministic route-coherence linter + CI gate (#2151)

krisztiankoos:
## What

Adds `scripts/quality/check_route_coherence.py` (+ CI wire) to catch the **cross-module / route-integrity** defect class the per-module QA pipeline is structurally blind to. Every existing gate is per-module (density, 7-dim rubric, cross-family review) or per-line (site-health checks links *resolve* — a backward `Next Module` link still resolves, so it passes). Nothing checked the relationships *between* modules and their section route. Closes the mechanical subset of #2151.

Three hand-audits (ai-ml / platform / on-prem) proved this pattern is site-wide, and that GLM's section audits miss intra-section chain breaks (e.g. planning 1.4→1.5). This fixes the *class*.

## Checks

| rule | level | what |
|---|---|---|
| `next-broken` | ERROR | Next-Module link target does not resolve |
| `next-backward` | ERROR | Next link points to an earlier / self module in the same section |
| `next-skip` | warning | Next skips the immediate successor / leaves the section early |
| `next-missing` | warning | non-terminal module has no Next link |
| `terminus-orphan` | warning | last module of a section has no forward Next link |
| `count-drift` | warning | index table lists ≠ on-disk `module-*.md` count (names the missing ones) |
| `marker-drift` / `marker-dual` | warning | delegates to `normalize_complexity_markers` (#2217) |

## How it stays low-FP (this took several iterations)

- **Link resolution mirrors `check_site_health`** — tries file-relative then the Starlight directory model (`foo/bar.md` → URL `foo/bar/`), so "resolves" means the same thing. This alone cut a naive 485 false `next-broken` → **0**.
- **`./` resolves to the section index, not self** (file-relative candidate ordered first) — killed self-link false positives.
- **Nav guards:** skips `Previous` / `Related` / `back to` links (label before OR *inside* the `[...]` text), ignores `## Next Steps` content sections, takes the first genuine forward link only.
- **`count-drift` matches absolute AND relative module links** and filters to real in-section files (cut 6 → 1, and the 1 remaining is real: `ml-platforms` index lists 8 of 11).

## Baseline ratchet

`scripts/quality/route_coherence_baseline.json` records the **6 known `next-backward` violations** on `main` today. CI fails only on **NEW** ones — so the gate prevents regressions immediately while the existing inventory drains as the curriculum lane fixes links (refresh with `--update-baseline`). The 6 are the seed for curriculum follow-up; the clearest real defects are `ai-infrastructure/module-1.6` and `module-1.8` (both "Continue to [earlier module 1.3]").

## Verification

- 17 unit tests (`scripts/quality/test_check_route_coherence.py`): number parsing (numeric not lexical — 0.10 > 0.2), resolution, nav extraction/skip, backward + broken integration.
- ruff clean; `check_route_coherence.py` CI-mode exits 0 on current main (0 new errors beyond baseline).
- Composes with #2217: `marker-drift` check finds **0 drift** (all markers canonical after the normalizer merge).
- Wired into `.github/workflows/link-check.yml` (every PR, stdlib-only, `contents: read`, **zizmor-clean**).

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -49,3 +49,13 @@ jobs:
         # as raw code in the browser). This stdlib-only lint closes that gap
         # deterministically. Author fix: `python scripts/ci/check_mermaid_labels.py --fix`.
         run: python scripts/ci/check_mermaid_labels.py
+
+      - name: Check route coherence
+        # site-health checks that links RESOLVE; this checks the relationships
+        # BETWEEN modules and their section route (a backward "Next Module" link
+        # still resolves, so site-health passes it). Fails only on NEW backward /
+        # broken Next-Module links beyond the committed baseline (ratchet) — the
+        # existing inventory drains as the curriculum lane fixes it. Warnings
+        # (skip / terminus-orphan / count-drift) report but do not fail. stdlib-only.
+        # Refresh the baseline: python scripts/quality/check_route_coherence.py --update-baseline
+        run: python scripts/quality/check_route_coherence.py
--- a/scripts/quality/check_route_coherence.py
+++ b/scripts/quality/check_route_coherence.py
@@ -1,0 +1,399 @@
+#!/usr/bin/env python3
+"""Deterministic route-coherence linter (issue #2151).
+
+Catches the cross-module / route-integrity defect class the per-module QA pipeline
+is structurally blind to (a backward "Next Module" link still *resolves*, so
+site-health passes it). Complements `check_site_health.py` (which checks that links
+resolve) by checking the *relationships between* modules and their section route.
+
+Checks
+------
+ERROR (fail CI):
+  * next-broken     — a module's "Next Module" link target does not resolve to a
+                      known page.
+  * next-backward   — the Next link points to an EARLIER module in the same section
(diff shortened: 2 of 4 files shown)


krisztiankoos:
## Cross-family review (composer-2.5) → APPROVE (R1)

Verified on the live corpus: resolution correct (`../module-x`, `./`, `/abs`), extraction low-FP (Related/Previous/"Next Steps" all correctly skipped), ratchet safe (main → 6 baselined, exit 0; a new backward link → exit 1), direction logic sound (0.10 > 0.2), and **CI will not block PRs on main**. 17 tests pass, ruff clean, zizmor clean.

Two non-blocking P2 hardening items tracked as follow-up (not triggered on the current corpus — reviewer confirmed 0 live false-positives):
1. Resolve relative links from the source's path-derived slug rather than its frontmatter `slug:` (matters only if a custom `slug:` ever diverges from the directory).
2. When both resolution candidates are valid, prefer the same-section `module-*` sibling.

Reviewer verdict: "Ship; track P2 slug-base / dual-candidate hardening as follow-up if custom slugs proliferate."

Issue #2310, opened 2026-09-04
[Content upgrade G02] Require revision-bound source evidence for acceptance


krisztiankoos:
Parent: #2274, follow-up to #2305/#2308 within epic #2272.

## Problem

Stopping injection from setting citations_verified prevents a false pass, but does not supply a trustworthy promotion path. The existing verifier's ok means execution completed, failing_count excludes UNREADABLE, and an empty claim set can report no failures. A durable receipt must prove coverage and bind the actual independent review to the exact page and claim/source revision.

## Required design and implementation

- First inspect the current claim extraction, seed, verification output, inventory receipt and readiness API schemas; publish an exact-path split plan before implementation. Do not invent a parallel status system or silently replace existing states.
- Define a schema-versioned receipt tied to content digest, claim/seed digest, source snapshot or retrieval locators, verifier identity/family and completed-run identity. Define what material-claim coverage means and how omitted claims are detected; an unchecked list is not a coverage proof.
- Fail closed for absent/malformed/schema-invalid receipts, stale page or seed revisions, empty coverage, unsupported/contradicted/unreadable claims, failed/partial provider runs, and a reviewer from the author family where independence is required. Preserve the distinction between tool execution, source reachability and semantic support.
- Only a complete, applicable, positive receipt can allow source acceptance. A manual author-set frontmatter bit or injection success cannot bypass the gate. Preserve citation-free drafting and independent review before source insertion.
- Add evidence-bound API/status reporting and a controlled promotion operation. Do not bulk promote historical pages or rewrite generated state without a separately reviewed execution packet.
- Test adversarial stale/missing/partial/unreadable/empty receipts and the valid complete path with deterministic fixtures; do not imply fixture tests prove real source support. Document migration/backfill and rollback boundaries.

## Delivery

Each PR owns fewer than20 files and at most200 aggregate changed lines unless a reviewed split packet sets another budget. Use worktrees, preserve unrelated state, full applicable Ruff/pipeline checks, independent-family code review and exact revision evidence. No production backfill or paid verification run is implied by implementing the receipt infrastructure. Close only when the promotion path and failure cases are demonstrated; whole-curriculum claim verification remains #2275.

krisztiankoos:
# #2310 implementation split — source acceptance

Current inspection base: maina090727e7. This is a proposed integration plan, not a completed gate or acceptance of any page.

## Existing seams and limits

[...]

PR #2330, opened 2026-09-05
fix(citations): validate cached source bytes before verification (#2328)

### LINKED ISSUE #2328: Validate cached source bytes before semantic citation review
Next bounded #2310 packet after merged#2323 producer provenance. The legacy semantic verifier currently reads cached text without metadata and silently fetches on a miss despite its offline claim. Bind what the verifier sees to validated cache bytes before any provider dispatch.

Own scripts/verify_citations.py and new tests/test_verify_citations.py only; <=250aggregate lines to cover the consumer plus adversarial integration tests. No caller/publisher/schema migration outside these files. Read cache text bytes and companion JSON without network. Reject missing/legacy/malformed metadata, failedHTTPstatus, issues/truncated results, digest/bytecount mismatch, invalidUTF8 and missing attempted-fetch timestamp as UNREADABLE with an explicit reason and zero dispatch. Old records remain unknown; do not silently refetch or manufacture provenance. The existing shorttextcheck remains. A matched hash proves byte correspondence, not source authenticity, truth, freshness or whole-page acceptance.

For dispatchable claims, carry the source URL/finalURL/textSHA256/bytecount/fetch-attempt time into the verdict record, plus digest of the exact composed verifier prompt and whether/howmuch source text was clipped by MAX_PAGE_CHARS. Hash the bytes actually used, not a later cache reread. Preserve this provenance on provider/parse failure too. Existing firstURL selection, claimselection, verifieridentity/outputattestation, page/seed/poolrevision binding, verdictschema validation and publication semantics are separate remaining#2310 requirements; do not claim fullgateacceptance or change ok/failing_count semantics in this packet.

Tests: fake cache+dispatch only; valid bytes reach provider and return exact provenance; missing/legacy/tampered/invalid/failed/truncated/short cases do not dispatch; providerfailure retains inputidentity; prompt clipping is represented accurately. No real network/provider calls. Resolve baselineRuff3importfindings in touchedscript, preserve unrelatedbehavior.176pipeline/Ruff/focusedtests/diffcheck and independent-familyreview beforemerge. NoAstrobuild requiredpurePython.

krisztiankoos:
Fixes #2328; advances #2310 within #2272.

The semantic citation verifier previously read mutable cached text without validating its metadata and could silently fetch a missing source. It now rejects missing, legacy, malformed, failed, truncated or mismatched cache records before provider dispatch. Dispatchable verdicts carry the exact source-byte identity and composed-prompt digest, including source clipping; provider and parse failures preserve that input provenance.

This is cache-integrity evidence. It does not establish source authenticity, freshness, claim coverage, actual reviewer identity or publication acceptance. Existing claim selection, first-URL choice and `ok`/`failing_count` semantics remain unchanged under the remaining #2310 work.

Validation: 73 focused verifier/fetcher/quality-stage tests; 176 pipeline tests; Ruff on both changed Python files; `git diff --check`. The independent review requested an exact failure-path prompt-digest assertion; that correction passed all 18 verifier tests and independent re-review at `c0e191038430c5907ff63ecdddcf7e458fba9d0d`. No live provider/source fetch or cache migration. No local Astro build is required for this Python-only change. Review comments are posted; all required CI must pass before merge.

Scope: two files, 223 additions + 26 deletions = 249 aggregate lines, within #2328's explicit 250-line budget. Primary checkout remains clean on `main`; runtime test artifact excluded.

### DIFF
--- a/scripts/verify_citations.py
+++ b/scripts/verify_citations.py
@@ -6,11 +6,9 @@ cached URL page text and returns a per-claim verdict:
 
     SUPPORTED / UNSUPPORTED / CONTRADICTED / UNREADABLE
 
-This gives 2-LLM independence: Codex researched the claim→URL pairing;
-a different LLM (Gemini by default) now reads the page and decides
-whether the URL actually backs the claim. Session 5 identified this
-as the missing high-trust gate. The fetcher's on-disk cache means the
-verifier never touches the network.
+Cache inputs are read offline without refetching source content. Semantic
+review may call the configured provider; this cache-integrity gate does not
+establish provider identity or author/reviewer independence.
 
 Usage:
     python scripts/verify_citations.py <module-key>
@@ -22,20 +20,26 @@ from __future__ import annotations
 
 import argparse
 import datetime as _dt
+import hashlib
 import json
 import sys
 from pathlib import Path
 from typing import Any
 
 sys.path.insert(0, str(Path(__file__).resolve().parent))
-from citation_backfill import (  # type: ignore  # noqa: E402
-    CITED_DISPOSITIONS, DOCS_ROOT, REPO_ROOT,
-    dispatch_codex, dispatch_agy, parse_agent_response,
-    resolve_claim_source_urls, resolve_module_path, seed_path_for,
+from citation_backfill import (  # type: ignore
+    CITED_DISPOSITIONS,
+    DOCS_ROOT,
+    REPO_ROOT,
+    dispatch_agy,
+    dispatch_codex,
     load_section_pool,
+    parse_agent_response,
+    resolve_claim_source_urls,
+    resolve_module_path,
+    seed_path_for,
 )
-from fetch_citation import cached_text_path, fetch  # type: ignore  # noqa: E402
-
+from fetch_citation import cached_text_path  # type: ignore
 
 VERDICT_DIR = REPO_ROOT / ".pipeline" / "citation-verdicts"
 MAX_PAGE_CHARS = 40_000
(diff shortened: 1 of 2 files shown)


krisztiankoos:
### Review Report

**Status:** REQUEST_CHANGES
**HEAD:** `39c35c5903241a2d8f39255f55efee4831494420`

#### Review Limits
*   **Deferred Scope (Not evaluated):** Actual model/family/run attestation, page/seed/pool revision binding, independent full-page claim coverage, output verdict schema validation, acceptance validator, and publication integration.
*   **Exclusions:** Cache hash matching is not evaluated as proof of source authenticity, truth, freshness, or whole-page acceptance. Provider aliases are not evaluated as identity evidence.

---

### Findings

#### 1. Introduced Defect: Incomplete Test Assertion
In `tests/test_verify_citations.py`, the final assertion in `test_dispatch_failures_keep_validated_provenance` is an incomplete, dangling expression:

```python
    assert result["source_url"] == URL and result["text_sha256"] == hashlib.sha256(PAGE_BYTES).hexdigest() and result["prompt_sha256"]
```

Because this statement ends abruptly with `and result["prompt_sha256"]`, Python evaluates the final condition strictly for "truthiness". While the test passes because the prompt hash is a non-empty string, it fails to perform the intended equality check against the expected prompt digest.

#### 2. Implementation Notes (Compliant / Passing)
*   **Cache Offline Enforcement:** The implementation properly severs the fetcher dependency (`fetch(url)` is removed). Missing metadata, mismatched hashes, and HTTP 500 errors natively trigger `UNREADABLE` without attempting a network refetch.
*   **Provenance Durability:** The provenance footprint (`prompt_sha256`, `text_bytes`, `text_sha256`, clipping data) is populated in a pre-dispatch `base` dict. It is correctly embedded into the result unconditionally—even when `dispatch_agy` fails or `parse_agent_response` raises exceptions.
*   **Encoding & Type Boundaries:** Explicit type checks (including `isinstance(..., bool)` prevention) and robust decode barriers (`UnicodeDecodeError`, `json.JSONDecodeError`) guarantee that malformed or tampered inputs are caught as cache metadata errors before they can poison prompt formation.
*   **Byte vs Character Handling:** Truncation and clipping logic correctly leverages character length (`len(page_text)`) for prompt limits (`MAX_PAGE_CHARS`), while disk validation strictly uses raw payload counts (`text_bytes` and `hashlib.sha256(body)`).
*   **Compatibility:** Upstream callers receive the exact same base keys (claim_id, url, verdict, reason) plus the new provenance expansions. Overwriting the LLM response via `.update(base)` prevents spoofing of these core tracking fields.

krisztiankoos:
I have reviewed the changes in the provided worktree at the current HEAD.

**Finding Resolution:**
The prior finding is **resolved**. The diff confirms that the compound truthiness assertion in `tests/test_verify_citations.py` has been split. The test `test_dispatch_failures_keep_validated_provenance` now explicitly extracts the `prompt` passed to the mock dispatch (`dispatch.call_args.args[0]`) and asserts strict equality between `result["prompt_sha256"]` and the actual UTF-8 encoded SHA256 digest of that prompt. I also confirmed that Ruff is clean and the 18 focused tests pass successfully.

**Verdict:**
**APPROVE** for HEAD `c0e191038430c5907ff63ecdddcf7e458fba9d0d`.

**Limitations & Context:**
*   This review validates the specific test assertion fix introduced in `c0e191038`. No implementation code was modified since the previous review (`39c35c590`).
*   Prior review coverage applies: Cache offline enforcement, input provenance, UTF8/hash validation, and clipping are covered. 
*   Existing deferred #2310 actual identity/coverage/revision/acceptance/publication requirements remain explicitly deferred. 
*   No live provider/source verification or blanket proof of every malformed-input case is claimed by this approval.

Issue #2279, opened 2026-09-04
[Content upgrade T02] Upgrade every Linux section with verified system behavior and practical diagnosis


krisztiankoos:
Parent epic: #2272

Type: workstream. Accountable lead: the Codex content-upgrade task.

## Scope and required change

Cover all 49 pages/37 modules in foundations, operations and security, including container primitives. Separate Linux kernel behavior, distribution/tool differences, and macOS host limitations. Validate privilege, namespace, network and cleanup boundaries.

Owned scope (no unrelated edits):

- `src/content/docs/linux/`

## Evidence and limits

Namespaces covers many namespace types plus a long privileged lab in a 35–45 minute banner; systematic troubleshooting reveals the injected root cause.

## Dependencies and coordination

#2273, #2274

Dependencies are per affected page/section: satisfy the prerequisite evidence for that packet, not completion of an entire umbrella workstream. Parent workstreams coordinate scope; they must not create circular wait-for-parent-completion gates. Check live leases/PRs and reserve disjoint paths before work.

## Acceptance criteria

- [ ] Every section has path-bounded review/revision packets and official manual/kernel/distribution sources.
- [ ] Privileged examples identify environment and reset procedure; exercise outputs are actually observed.
- [ ] Diagnosis tasks require evidence rather than copying a supplied cause.
- [ ] Learning progression, workload, assessments and translated counterparts are reconciled.

## Delivery and factual standard

- No invented facts, citations, incidents, quotations, results or learner feedback. Verify exact claim support from authoritative sources; record locators, dates, versions and uncertainties. Clearly label reproducible synthetic exercises. No forced mystery or gamification.
- Workstream issues must create exact-path execution packets covering their full scope; a representative sample does not complete a portfolio. Preserve good content with a reviewed retain disposition.
- Use worktrees, apply_patch and scoped PRs. Default budget: at most 200 net changed lines and fewer than 20 files; split before exceeding unless an issue explicitly sets another reviewed budget. Preserve unrelated/user changes and generated runtime state.
- Run applicable source, lab, translation, lint/test/build/link/render checks. State unexecuted checks and why. A heuristic pass or reachable URL is not factual or lab verification.
- Cross-family review and all required gates must pass before merge. Close only with exact revision, review and delivery evidence; no blanket status promotion.

krisztiankoos:
[...]

PR #2333, opened 2026-09-05
fix(linux): ground storage diagnosis in observed write probes (#2331)

### LINKED ISSUE #2331: [Content upgrade T02] Replace the Linux lab’s unsupported disk-full diagnosis with measured evidence
Parent #2279; complete upgrade epic #2272.

## Concrete defect

In `src/content/docs/linux/operations/troubleshooting/module-6.1-systematic-troubleshooting.md`, Hands-On Exercise Part 2 writes 100 MiB to a fixed `/tmp/bigfile`, invents an application write-failure message, and prints a disk-full root cause without observing a failed write. The allocation does not establish that the filesystem is full; the fixed name can overwrite an unrelated file. The exercise teaches the opposite of its hypothesis-testing outcome.

## Bounded correction

Own that English module only, Part 2 and its directly related success criterion/source link; <=200 aggregate lines. Replace the pretend outage with a clearly labeled, guided storage experiment that tests whether a small controlled allocation actually supports a disk-full hypothesis. Use a unique temporary directory, a bounded small payload, an actual baseline and post-allocation write probe, visible exit/error evidence, and cleanup limited to resources this run created. Do not fill a host filesystem, change permissions on unrelated paths, or claim a simulated alert is observed evidence.

Ask the learner to predict, record observations, distinguish filesystem-wide capacity from this directory's allocation, reject unsupported conclusions and explain what further evidence a genuine write failure would require. This is guided practice, not a blind incident assessment or proof of service-troubleshooting mastery. No forced mystery or hidden fabricated incident. Expected outcomes must be distinguished from results actually observed during validation; an unexpected failure remains a finding to investigate.

## Verification

Execute the exact new fenced commands in an isolated Linux environment, recording version/context, write outcomes, bounded resource use and cleanup. Exercise an early failure path and ensure it cannot fall through into unrelated deletion. Parse the affected shell fences; run applicable pipeline/build/render checks and independent-family review. No production incident or learner-performance claim is authorized by a fixture run. Preserve the rest of the module and record other findings as later #2279 packets.

Method source inspected: [Google SRE, Effective Troubleshooting](https://sre.google/sre-book/effective-troubleshooting/), Theory and Common Pitfalls: compare observations to predictions, use controlled changes, and distinguish correlation from causation. Verify command behavior with appropriate upstream manuals and actual execution. Initial GNU coreutils manual requests timed out; inaccessible URLs do not count as checked sources.

krisztiankoos:
Fixes #2331; advances Linux workstream #2279 in epic #2272.

Part 2 of systematic troubleshooting previously wrote a fixed `/tmp/bigfile`, printed an invented application error and declared a disk-full root cause without observing failed I/O. It now uses a unique temporary directory and a bounded 256 KiB allocation, with baseline/post-allocation write probes, actual exit status/stderr, a capacity guard and cleanup. Learners predict outcomes and explain which evidence supports or rejects the hypothesis. The success criterion describes guided practice, not independent incident mastery.

Validation: exact lesson fence executed as non-root in isolated arm64 Linux (Bash 5.2.15, coreutils 9.1), no network/host mounts and read-only root. The normal 16 MiB tmpfs case passed both writes/allocation/cleanup; missing TMPDIR stopped before writes; a 512 KiB tmpfs refused allocation and cleaned up. Source-page SHA256 `4a0ddae0a1a4eba75f72c92f8643cf19390bdb7c14c6a6015d005e2be3d86954`; image ID `sha256:4613778f3cfcd10e615029370f5786704559103cf27bef934597ba562b269661`. No production incident or learner-effectiveness claim follows from these runs.

Shell syntax, 176 pipeline tests, local 2,169-page build, rendered code/links and `git diff --check` passed. The Google SRE methodology and upstream coreutils manuals were inspected. Independent-family review is in progress and required before merge. Other exercise parts, whole-module acceptance and translation are separate remaining work.

Scope: one module, 99 additions + 22 deletions = 121 aggregate lines within the 200-line budget. Primary checkout remains clean on `main`; ignored validation logs, receipts and generated artifacts are excluded.

### DIFF
--- a/src/content/docs/linux/operations/troubleshooting/module-6.1-systematic-troubleshooting.md
+++ b/src/content/docs/linux/operations/troubleshooting/module-6.1-systematic-troubleshooting.md
@@ -497,11 +497,11 @@ The timeline should record the first known impact, key observations, hypotheses 
 
 ### Practicing Systematic Troubleshooting
 
-**Objective**: Apply troubleshooting methodology to a simulated problem.
+**Objective**: Apply troubleshooting methodology to a bounded storage experiment.
 
 **Environment**: Any Linux system
 
-This exercise is intentionally local and safe. You will collect a baseline, simulate a small disk-space symptom in `/tmp`, test hypotheses, inspect recent changes, and document the session. The commands should run on a normal Linux workstation or lab VM, though service names and permissions can vary by distribution.
+This exercise is intentionally local and bounded. You will collect a baseline, test a small storage hypothesis with real write probes, inspect recent changes, and document the session. The commands should run on a normal Linux workstation or lab VM, though service names and permissions can vary by distribution.
 
 #### Part 1: Initial Triage Script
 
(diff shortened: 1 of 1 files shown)


krisztiankoos:
I have completed the read-only independent review of the specified worktree based on your criteria. 

**Review Result:** APPROVE

**Exact HEAD:** `b725683a5`

### Material Findings

**1. Source Verification & External Calls Limit**
In strict compliance with the instruction to make no external provider calls or delegations, I did not fetch the external `man7.org` or `sre.google` URLs over the network. Instead, I ran `man mktemp`, `man df`, and `man dd` in the local terminal to verify the flags (`-d`, `-P`, `-k`, `bs`, `count`) directly against the GNU coreutils manuals on the host. I evaluated the pedagogical correctness using internal knowledge of the Google SRE Book's "Effective Troubleshooting" chapter. I did not invent any external verification.

**2. Cleanup Safety**
The implementation provides excellent safety guarantees. 
- The script executes in an isolated child shell (`bash <<'EOF'`), protecting the learner's parent shell state.
- `probe_dir` is initialized empty and dynamically assigned via `mktemp -d`.
- The `trap cleanup EXIT` uses defensive checks `[ -n "$probe_dir" ] && [ -d "$probe_dir" ]` before running `rm -rf -- "$probe_dir"`. This guarantees it will only ever delete the single unique directory created for this run and never a system path, even if the script aborts unexpectedly.

**3. Failure Propagation**
The error handling is exceptionally robust.
- The script captures `$?` immediately after critical commands (`setup_rc`, `baseline_rc`, `allocation_rc`, `post_rc`).
- If any probe or allocation step fails, it emits a specific diagnostic string to `stderr` and immediately calls `exit "$rc"`. 
- The `cleanup` function perfectly preserves this original exit code while executing the teardown, or elevates the exit code to `1` if the teardown itself fails. 

**4. Capacity Guard**
The script effectively prevents exacerbating a real storage incident:
- It correctly extracts the available KiB using `df -Pk` and `awk 'NR == 2 {print $4}'`.
- The `case` block safely handles unreadable/non-numeric edge cases.
- It correctly enforces a `>= 1024` KiB (1 MiB) limit before attempting the 256 KiB allocation, exiting gracefully with code `2` if there is insufficient space.

**5. Pedagogical Correctness & Wording**
The script perfectly embodies the SRE methodology of bounded testing and evidence observation:
- The distinction between filesystem capacity (`df -P`) and directory allocation (`du -sh`) is well-illustrated. 
- The "Solution notes" clearly emphasize that this bounded test provides evidence *against* a local write failure, but does *not* blindly prove that the wider service or filesystem is perfectly healthy (e.g. inode exhaustion, permissions, quotas). 
- It actively guides the learner to rely on the actual logged outputs (`dd` stderr, exit codes) rather than jumping to a pre-conceived "root cause", replacing the previous artificial "Simulate: Fill up /tmp" approach entirely.

Issue #2358, opened 2026-09-05
[Content upgrade] Validate document route-set input before link checks


krisztiankoos:
Child of #2346; full upgrade #2272. The native URL resolver (#2348) and opt-in Starlight document manifest (#2357) are merged. Implement the first consumer boundary without pretending it is a whole-site link gate.

Add a pure validated document-route-set adapter and meaningful fixtures. Input is the schema1 `starlight-document-routes` manifest; output is the document target path set (including real fallback targets) and exactly one non-fallback canonical URL per concrete source. Reject malformed or unsupported config, URL/path mismatches, unsafe source paths, inconsistent source hashes, duplicate target paths, missing primary source mappings, and multiple primary mappings. Preserve fallback metadata. Do not reconstruct a route from a source filename or silently choose a primary record.

Explicit limits: source hashes are validated as metadata consistency, not compared to current filesystem bytes in this pure adapter. Redirect configuration is not accepted as a target; custom Astro routes are not invented or hardcoded. Those require separate same-build rendered proof and bounded redirect resolution before the final gate. A later consumer will check actual source bytes and source completeness against the controlled build.

Two owned paths: `scripts/quality/document-route-set.mjs` and `scripts/quality/document-route-set.test.mjs`. Default <=200 aggregate changed lines; split rather than compress tests. Tests must include primary+UK fallback, a concrete Ukrainian primary, same-source conflicting hash, missing/two primary mappings, URL/path normalization mismatch, non-origin/cross-origin routes, unsupported schema/config, and duplicate canonical paths. Use native URL behavior and serialized path evidence.

No legacy gate replacement, baselines, or acceptance of known missing hrefs in this packet. Required Node tests, pipeline checks, diff, independent Google review, CI, clean primary main; no Astro build needed unless runtime configuration or published content changes. Parent #2346 remains open until full consumer integration and failure reporting are implemented.

krisztiankoos:
Delivered in #2361 after all CI and independent Google review1788584814. The pure adapter validates schema/config, serialized same-origin paths and base containment, source metadata/hash consistency, duplicate targets, and exactly one primary source mapping. Real UK fallback targets stay distinct from authored sources.

[...]

Issue #2346, opened 2026-09-05
[Content upgrade] Check canonical browser routes for EN and UK links


krisztiankoos:
Parent #2293; related #2343, #2344 and full upgrade #2272.

## Verified false-pass mechanism

Read-only scout at main 2829a2e0d found that scripts/check_site_health.py:207–210 skips UK pages, while :234–250 accepts EN relative targets resolved against the Markdown parent directory. A temporary two-page EN/UK fixture with the broken next-lesson href reported one link checked and zero potentially broken: English passed the source-relative fallback, Ukrainian was omitted.

The site's trailingSlash: always configuration makes browser resolution different. From canonical /ai/foundations/module-1.1-what-is-ai/, ./module-1.2-what-are-llms/ resolves beneath the current module directory, not to its sibling. PR #2344 corrects one English link, with UK coordination in draft #2342, but does not fix the checker.

scripts/quality/check_route_coherence.py:181–187 repeats parent-first candidates; :197–200 skips UK. Its test at scripts/quality/test_check_route_coherence.py:63–66 codifies that fallback. scripts/check_links.py uses urljoin but is not the gate invoked by npm run check / .github/workflows/link-check.yml.

## Required change

Design one deterministic canonical-route resolver shared or consistently used by the affected source gates. Derive routes from actual slug/index/base/trailing-slash rules, resolve hrefs with browser URL semantics, and compare with the route manifest for both EN and UK. Preserve query/fragment handling and make unsupported route forms explicit. Do not silently prefer source-file siblings or omit Ukrainian links.

Add regression fixtures for a broken ./sibling link, correct ../sibling link, UK relative links, index routes, explicit slugs, root-relative links and fragments. Show that the original false-pass fixture fails and the corrected links pass. Report existing site failures separately with exact paths; do not weaken assertions, hide failures or bulk-edit unrelated curriculum to make the checker green.

Use an isolated worktree, default 200 aggregate-line / fewer-than-20-file budget; post a measured split plan if a coherent resolver/integration requires more. Required focused tests, Ruff for changed Python, pipeline checks and independent-family review apply. This issue improves route verification, not factual or translation acceptance.

krisztiankoos:
Lead-reviewed split plan for #2346. The stricter resolver must expose actual failures; the existing route baseline will not be expanded to hide them.

[...]

PR #2361, opened 2026-09-05
feat(quality): validate document route-set metadata (#2358)

krisztiankoos:
Validate the document manifest before a link checker uses its routes. The adapter rejects malformed configuration, inconsistent URL/path pairs, duplicate paths, conflicting source hashes, and missing or ambiguous primary source mappings. Real Ukrainian fallback routes remain valid document targets while authored sources use their unique primary URL.

Refs #2358 and #2346. This is a pure metadata adapter: it does not verify current source bytes/completeness, add custom Astro routes, resolve redirects, or replace the legacy link gate. Redirect metadata never becomes an accepted target through this adapter.

Validation: five focused Node test groups; 176 pipeline tests; acceptance of the prior controlled manifest's2,166 targets/1,639 sources/527 fallbacks, pinned by manifest hash; diff check; independent Google review posted separately. No build needed for these two pure-code files. 156 added lines, no generated artifacts, clean primary main. Parent #2346 stays open for full integration.

### DIFF
--- a/scripts/quality/document-route-set.mjs
+++ b/scripts/quality/document-route-set.mjs
@@ -1,0 +1,94 @@
+export class UnsupportedDocumentRouteManifest extends TypeError {}
+
+const fail = (message) => { throw new UnsupportedDocumentRouteManifest(message); };
+const object = (value, label) => {
+  if (!value || typeof value !== 'object' || Array.isArray(value)) fail(`${label} must be an object`);
+  return value;
+};
+const text = (value, label) => {
+  if (typeof value !== 'string' || !value) fail(`${label} must be non-empty text`);
+  return value;
+};
+const optionalText = (value, label) => {
+  if (value !== null && (typeof value !== 'string' || !value)) fail(`${label} must be non-empty text or null`);
+  return value;
+};
+
+function originFor(config) {
+  const site = text(config.site, 'config.site');
+  let url;
+  try { url = new URL(site); } catch { fail('config.site must be a URL'); }
+  if (!['http:', 'https:'].includes(url.protocol) || url.username || url.password || url.search || url.hash) fail('config.site must be an HTTP(S) URL without credentials, query, or fragment');
+  if (typeof config.base !== 'string' || !config.base.startsWith('/') || config.base.startsWith('//') || config.base.includes('?') || config.base.includes('#')) fail('config.base must be a root-relative path');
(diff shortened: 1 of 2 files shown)


krisztiankoos:
Google review smart-agy-review-1788584814 of head0aa571e86 passed. Lead also exercised the real pinned manifest; no fresh source-drift acceptance is claimed.

[smart] agent=agy task_class=review model=gemini-3.1-pro-high mode=read-only timeout=180s
[smart] cwd=/Users/krisztiankoos/projects/kubedojo/.worktrees/content-document-route-set
[smart] task_id=smart-agy-review-1788584814
======================================================================
OK: True  |  elapsed: 40s  |  resp_chars: 1876
response_path: logs/dispatch_responses/smart-agy-review-1788584814.txt
======================================================================
**PASS**

The read-only review of `HEAD` against `main` for `content-document-route-set` yields no material findings. The code correctly fulfills the pure adapter requirements. 

**Evidence:**
- **Test execution**: I successfully ran the 5 focused Node test suites natively (`node --test scripts/quality/document-route-set.test.mjs`). All tests passed in ~40ms, verifying that the schema validation and edge-case rejections function as designed.
- **Producer compatibility & Schema validation**: `document-route-set.mjs` handles malformed inputs rigorously. It uses strict explicit checks (e.g. `typeof` checks rejecting `undefined` where explicit `null` is required like `locale`/`lang`, exact `origin`/`pathname` matching against parsed `URL` properties) to ensure adherence to the expected Starlight schema, rejecting non-serialized, cross-origin URLs, or malformed/unsupported config settings.
- **Mapping constraints & Fallback handling**: The module cleanly iterates over the source mappings, correctly enforcing exactly one primary route per source while preserving UK fallback paths as requested (`fallbacks: routes.filter(...)`). It safely catches hash conflicts (verifying that all entries tied to a source string resolve to the exact same SHA-256 footprint).
- **Base containment logic**: The verification of `pathname` strictly checks whether the string starts with the expected base-root segments (`!pathname.startsWith(root + '/')`).
- **Redirect exclusion**: Redirect metadata objects are thoroughly type-checked (e.g. ensuring `3xx` integer statuses) in `validateRedirects()`, but as requested, they are entirely bypassed when constructing `targetPaths` and `primaryBySource` indexes.

The adapter correctly confines itself to structural and semantic validation without expanding into filesystem verification, redirects resolution, or legacy baseline rewrites.

Issue #2418, opened 2026-09-05
fix(linux): bound namespace lab ownership and cleanup


krisztiankoos:
Parent #2279; coordinate #2276 and #2291; epic #2272.

Owned EN source: `src/content/docs/linux/foundations/container-primitives/module-2.1-namespaces.md`. At main `d11205d8058d003da330a99f3357879da91f713a`, the Part4 lab (around800–823) creates `/tmp/kd-mnt-lab`, mounts tmpfs there in a new mount namespace, exits, and runs an expected-failing `cat` on the host without removing the underlying directory. The earlier mount example similarly uses `/tmp/ns-mount-lab`. Reused fixed paths also need a pre-existence/ownership boundary before privileged writes or cleanup.

Lead inspected [util-linux unshare(1)](https://man7.org/linux/man-pages/man1/unshare.1.html), Description / mount namespace: since util-linux2.27 the command makes propagation private by default; shared propagation can affect other namespaces. This does not prove the existing lab was executed or cleaned up. Record exact Linux/kernel/util-linux versions and explicit propagation when validating.

Deliver a bounded EN lab repair first: use task-owned temporary paths, separate namespace-local mount teardown from host directory cleanup, make expected absence checks readable and exit-coded, and state the Linux/privilege/environment requirements. Validate normal exit and interrupted/failed setup in a disposable Linux environment; compare before/after mount, path and namespace state. Preserve user/pre-existing paths. Never run unreviewed privileged lesson commands on the Mac host or VPS.

Then reconcile workload and diagnosis tasks using actual timed execution or a clearly labeled estimate, not the scout's unsupported60–75minute replacement. Existing35–45minute claim is a review target, not a measured falsity. The missing time-namespace entry in overview loops needs an explicit scope decision or supported capability check, not an unguarded assumption that every kernel exposes it.

UK counterpart has substantial structural divergence and requires a separately budgeted full-body fidelity packet after the EN source is accepted; follow #2291/#2086. Do not infer truth from matching EN: IPC-sharing and other kernel/runtime claims need their own official-source verification. Existing citation flags and broad retain claims are not factual acceptance.

Acceptance: exact-path source mapping, bounded environment and cleanup, observed command/output receipts, proportionate workload, independent source/prose reviews, pipeline/build/render/CI/deployment evidence. Keep each PR below200aggregate lines/<20files unless an explicit split budget is approved in the issue. This issue does not accept the whole Linux track or a translated page.

[...]

PR #2432, opened 2026-09-05
fix(linux): retry shared-volume reader until file appears (#2418)

krisztiankoos:
The shared-volume reader exits if it opens the log before the writer creates it. Use BusyBox `tail -F` to retry, and explain the possible initial open error while distinguishing retry from startup ordering and application health. Keep the writer, two ordinary containers and volume layout unchanged.

Refs #2418, #2279; epic #2272. Other namespace claims and Ukrainian fidelity remain open.

Validation: inspected actual BusyBox1.36.1 arm64 help/image; controlled delayed-writer comparison (`Never`) and exact candidate manifest (`Always`) executed in the dedicated Kubernetes1.35 lab, with cleanup verified. Independent Grok SOURCE/DESIGN PASS before prose; separate Google PROSE PASS at `803a7718e8fcc8b7ff1de5c22d30a51e64ebd281`. All176 pipeline tests passed (4.257s), primary-root build passed (2169pages/39.52s), exact rendered manifest/explanation and old-reader removal verified. Two files,+24/-1 aggregate25lines, diff check clean; generated test artifact restored. CI/deployment/live pending. No frequency, general reliability or learner-outcome claim.

### DIFF
--- a/docs/content-upgrade/research/shared-volume-tail-evidence.md
+++ b/docs/content-upgrade/research/shared-volume-tail-evidence.md
@@ -1,0 +1,21 @@
+# Shared-volume reader: missing-file evidence
+
+Scoped evidence for #2418, parent #2279, epic #2272. Only the `shared-volume-demo` reader flag and its explanation are accepted by this packet; no IPC, signalling, whole-module or Ukrainian acceptance.
+
+## Inspected implementation and observed tests
+
+The tested `busybox:1.36` image resolved to digest `sha256:73aaf090f3d85aa34ee199857f03fa3a95c8ede2ffd4cc2cdb5b94e566b11662` on arm64. Its own `tail --help` identified BusyBox1.36.1 and described `-F` as retaining retry behavior. This binary help is the primary implementation evidence; an unversioned web option list did not describe the tested binary correctly. No GNU flag equivalence or rotation behavior is inferred.
+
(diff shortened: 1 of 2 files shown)


krisztiankoos:
Independent Google PROSE/exact-manifest **PASS** at `803a7718e8fcc8b7ff1de5c22d30a51e64ebd281`; runtime Gemini3.8FlashHigh, session `f14642bd-9451-4271-b788-aa92aa39d963`. Reviewer inspected both changed files and the implementation/pod receipts, including the exact default-Always candidate, actual date output, both running containers with zero restarts, and explicit cleanup provenance. No blockers in this scope.

Grok SOURCE/DESIGN review preceded prose, session `885aec63-a575-4095-85cd-83d4136feb48`, CursorGrok4.6High. Lead clarified its finding: the original comparison harness requested deletion, while lead verified absence separately; the receipt and note now distinguish that check. Candidate harness includes its own deletion wait and absence check. Final Google review covered the clarification.

Root176 tests/build/render passed. Render HTML SHA256 `b25c4efa799f80662af24f4c53582c092eea6693182e5a0949fe576ec1d1e6ce`. Evidence is bounded to tested BusyBox1.36.1 arm64 digest and controlled cases. No startup-order guarantee, default-restart-frequency result, rotation/GNU-equivalence claim, broader module or UK acceptance. CI/deployment/live checks still apply.

Issue #2478, opened 2026-09-09
feat(cka): validate isolated apiserver certificate renewal and rollback


krisztiankoos:
Parent #2280; full upgrade #2272. Follow-up to closed #2338 and source boundary card #2466. Completed containment #2468 and diagnostics #2476 do not establish practical renewal. Lifecycle-only #2477 provides an independently reviewed disposable fixture with actual create/inspect/delete evidence.

Own a dedicated apiserver-serving-certificate renewal and consumer-reload experiment on a fresh, uniquely owned kind/Kubernetes v1.35.0 cluster. Do not mutate the persistent fixture or shared/live infrastructure, change clocks, renew all certificates, or export node private keys. This demonstrates renewal/reload; it is not recovery from observed expiry.

Preserve the reviewed plan in #2338 comment 5609805429. Before implementation, measure and review the split for the complete transaction. Default <=200 aggregate changed lines and <20 files per PR; do not omit rollback to meet the budget. Publish no learner mutation recipe until the entire execution chain is independently accepted.

Required results:
- Identity-checked staging and exclusive node-side state/locking, private verified backups outside the static-pod watch directory, and recorded mutation phases before actions.
- Stop the consumer before renewal; renew only apiserver; establish changed public fingerprint, matching key/identity, unchanged manifest return, a newly running consumer and a trust-verified served certificate matching the new identity.
- Authenticated readiness plus marker reads and fresh writes after renewal and rollback. A later expiry timestamp is not required for same-second renewal.
- Coordinate signals, surviving child processes and explicit recovery inside the node. A cancelled Docker client is not proof that the mutator stopped. Stop the consumer before restoring originals; prove restored served identity and API function. Preserve unresolved state and refuse fixture deletion while recovery is unresolved.
- Exercise positive, partial failure and interruption boundaries. Label injected failures separately from real observations; deleting the fixture is not rollback proof.
- After full acceptance, a separate <=80-line lesson change may link the complete artifact and request public evidence. Preserve all diagrams and accepted scheduler exercise. Build/render and independent-family review required for publication.

Current evidence: no certificate renewal has executed. Complete node transaction is only a planning draft; line counts and remaining uncertainty must be reported honestly. This issue does not claim whole-module, hosted-lab or Ukrainian acceptance.

krisztiankoos:
[...]

PR #2480, opened 2026-09-09
feat(cka): add read-only certificate evidence primitives (#2478)

krisztiankoos:
Progresses #2478 with the first internal module from the measured split (comment 5609968780). Certificate renewal needs reliable observations before any recovery transaction can be enabled. This sourced Bash library validates explicit CRI identities, certificate/key pairing, the supplied CA and served certificate, and authenticated readiness/marker identity. Failures propagate under conditional callers; sourcing only defines functions.

No renewal, manifest movement, backups, API writes, or learner mutation instructions are added. The broader transaction remains disabled and unvalidated.

Validation:
- Fresh owned Kubernetes v1.35.0 fixture: positive CRI/certificate/TLS/API checks passed; wrong key, untrusted CA/served chain, and wrong marker UID were rejected. Certificate/four manifest/installed-client fingerprints stayed unchanged. Staged tools and fixture were removed; baseline clusters were preserved.
- 21 separately injected CRI cases passed under a conditional caller, including empty/malformed output, wrong component/namespace/state/ID, duplicate IDs, wrong client version, and command failure.
- Bash syntax, Ruff and whitespace checks passed; all 176 pipeline tests passed (2.459 seconds).
- Independent Google pre-execution review passed. Exact-head final review follows as a comment.

Two files, 193 added lines. No generated artifacts. Local Astro build is not applicable to this scripts-only packet. OpenSSL source references: [s_client](https://docs.openssl.org/3.0/man1/openssl-s_client/), [x509](https://docs.openssl.org/3.0/man1/openssl-x509/).

PR #2477, opened 2026-09-09
feat(cka): add isolated certificate fixture lifecycle (#2338)

krisztiankoos:
Adds an isolated disposable Kubernetes 1.35 fixture for the next certificate experiment in #2338. The helper creates a uniquely named, digest-pinned kind cluster, records its container and API identities, verifies authenticated marker access, and deletes it only after checking ownership. Private kubeconfig removal requires confirmed cluster/container absence.

This is lifecycle packet 1 of the split posted on #2338. It contains no certificate renewal or rollback transaction and adds no learner mutation recipe. Uncertain partial creation is preserved for reconciliation; client cancellation does not prove underlying work has stopped.

Validation:
- Ruff, syntax, and whitespace checks passed; pipeline suite: 176 tests passed.
- Actual fresh-cluster create, inspect, delete, and repeated delete passed. Kubernetes v1.35.0 readiness and marker write/read succeeded; original clusters, containers, and default kubeconfig were preserved; the owned private kubeconfig was removed.
- Eight separate injected function-level boundary cases passed: existing directory, concurrent invocation, insecure directory, cluster collision, wrong deletion identity, failed cleanup retaining credentials, confirmed absence, and repeated deletion. These are not real runtime-failure or interruption observations.
- Independent Google pre-execution review passed; exact-head final review follows as a PR comment.

One file, 197 added lines. No generated artifacts. Astro build is not applicable to this Python-only packet. Refs #2338; broader recovery acceptance remains open.

PR #2472, opened 2026-09-09
docs(platform): refresh OTel Collector landscape to v0.160.0 (#2469)

krisztiankoos:
Parent: #2294 (X03). Closes #2469.

## What

Currency-only refresh of `src/content/docs/platform/toolkits/observability-intelligence/observability/module-1.12-otel-collector-production.md`.

The June 2026 Landscape snapshot still named core **v1.60.0/v0.154.0** and contrib **v0.154.0** as current. Verified against GitHub releases on 2026-09-09:

| Distro | Previous pin | Current release | Published |
|---|---|---|---|
| Core (`opentelemetry-collector`) | v1.60.0/v0.154.0 (2026-06-08) | **v1.66.0/v0.160.0** | 2026-09-02 |
| Contrib (`opentelemetry-collector-contrib`) | v0.154.0 (2026-06-09) | **v0.160.0** | 2026-09-02 |
| Collector releases | — | **v0.160.0** | 2026-09-02 |

Sources:

- https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.160.0 (`v1.66.0/v0.160.0`)
- https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.160.0
- https://github.com/open-telemetry/opentelemetry-collector-releases/releases/tag/v0.160.0

## Changes

- Landscape snapshot date: `2026-06` → `2026-09`
- Core/contrib version claims + publish dates
- Sources “current release” links/labels for core and contrib
- Paired versioning note kept (still accurate: stable core `v1.x` / distribution `v0.1xx`)

## Out of scope

- No pedagogy, labs, or processor rewrite
- No Harbor/Cilium modules (#2467)
- No UK twin (path does not exist)

## Diff

1 file, 4 lines changed. Currency only.

[...]

PR #2481, opened 2026-09-09
feat(cka): add private certificate state foundation (#2478)

krisztiankoos:
Progresses #2478 with private initialization and locking for the later certificate transaction. The combined state/backup packet is narrowed to an immutable `created` state, explicit fixture/artifact identity, exclusive initialization, persistent lock inode, atomic state creation and guarded reopen/status. The host must independently verify Docker ownership.

There is no backup, stage transition, renewal or restoration entry point. Full recovery and surviving-child coordination remain unvalidated follow-up work. Identity is explicitly supplied, not signed.

Validation:
- Fresh owned Kubernetes v1.35.0 fixture: init/open/status passed; contention, duplicate initialization, wrong identity, altered state, unsafe permissions, symlink and changed artifact guards refused. These negative cases are injected.
- Separate fresh fixture: injected file-sync failure before rename refused initialization, left incomplete state unreopenable, closed the descriptor and refused collision retry.
- Both fixtures passed authenticated inspection, retained the API-server certificate/four manifest fingerprints, and were deleted with pre-existing clusters preserved. Teardown is not rollback evidence.
- Source-time caller/descriptor preservation and 18 explicit identity cases passed; Bash syntax, Ruff, whitespace and all 176 pipeline tests passed.
- Independent pre-execution review passed for the limited foundation. Exact-head final review follows as a comment.

Two files, 189 added lines. No generated artifacts. Local Astro build is not applicable to this scripts-only packet.

### DIFF
--- a/scripts/cka_certificate_state.sh
+++ b/scripts/cka_certificate_state.sh
@@ -1,0 +1,140 @@
+# Internal Linux Bash library: sourcing defines functions only.
+# Host verifies Docker identity independently; these functions do not prove it.
+# Requires root, /proc, jq, flock and GNU stat, sha256sum, mktemp, sync, mv.
+
+cka_cert_state_file() {
+  [[ $# == 1 ]] || return 1
+  [[ -f $1 && ! -L $1 ]] && [[ $(stat -c '%u:%a:%h' -- "$1") == 0:600:1 ]]
+}
+
+cka_cert_state_environment() {
+  local path mode
+  [[ $# == 0 && $EUID == 0 && -d /proc/$BASHPID/fd ]] || return 1
+  for path in / /var /var/lib; do
+    [[ -d $path && ! -L $path && $(stat -c %u -- "$path") == 0 ]] || return 1
+    mode=$(stat -c %a -- "$path") || return 1
+    (( (8#$mode & 0022) == 0 )) || return 1
+  done
+}
+
+cka_cert_state_directory() {
+  [[ $# == 1 ]] || return 1
+  [[ -d $1 && ! -L $1 ]] && [[ $(stat -c '%u:%a' -- "$1") == 0:700 ]]
+}
+
+cka_cert_state_expected() {
+  [[ $# == 1 ]] || return 1
+  jq -ces '
+    def hex: type=="string" and test("^[a-f0-9]{64}$");
+    if length==1 then .[0] else error("Expected one identity") end |
+    select(type=="object") |
+    select(keys==["artifact_hashes","docker_container_id","fixture_cluster",
+      "image_id","system_namespace_uid","transaction_id"]) |
+    select(.transaction_id | type=="string" and test("^[a-f0-9]{32}$")) |
+    select(.fixture_cluster | type=="string" and test("^cert-fixture-[a-f0-9]{32}$")) |
+    select(.docker_container_id | hex) |
+    select(.image_id | type=="string" and test("^sha256:[a-f0-9]{64}$")) |
+    select(.system_namespace_uid | type=="string" and
+      test("^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$")) |
+    select(.artifact_hashes | type=="object" and keys==["observe.sh","state.sh"] and
+      all(.[]; hex))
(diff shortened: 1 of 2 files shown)


krisztiankoos:
Independent Google-family exact-head review: **VERDICT PASS** at `8f550b5368da86de95c5ce89c1f87c722d221591` (Gemini via Cursor). Driver inspected and accepts the bounded result; no material findings.

The review covered both changed files and the two fresh-fixture receipts: actual initialization/reopen/status, injected contention and invalid-input/filesystem guards, and a separately injected pre-rename sync failure. Both fixtures retained the public certificate/manifest fingerprints and authenticated API function, then were removed with the baseline preserved. All 176 pipeline tests passed.

Identity is supplied explicitly, not signed; no duplicate-JSON-key rejection is claimed. Closing FD 9 avoids explicitly unlocking descendants, but surviving-child lifecycle behavior remains unvalidated. This accepts only the initialization/locking foundation: no backup, stage transitions, renewal, restoration or rollback acceptance. Fixture teardown is not rollback evidence.

krisztiankoos:
CF:PASS

**Head:** `8f550b5368da86de95c5ce89c1f87c722d221591` (packet 2 foundation / private state + locking for the measured #2478 split, comment 5609968780 — narrowed to immutable `created` only; backup custody deferred).

**Situational note:** Branch `codex/content-cka-cert-state` owns this slice of open #2478. Sibling open content work (`#2472` OTel, `#2471` Astro proof, `#2470` Harbor/Cilium, `#2270` header) does not collide. Lifecycle `#2477` and read-only observe `#2480` are already on main; no second `codex/*` twin for this state library. Live CKA learner pages are untouched (correct for this packet).

**Learner / editor walk**
- **Scope claim holds:** two files, 189 added lines — inside the ≤200 / <20 default. Sourcing defines functions only; no backup, stage transition, renewal, restoration, or learner mutation entry point. Matches the PR’s honest narrowing of measured packet 2 (exclusive dirs/locks + atomic `created` state); verified backup custody and manifest/pair checks remain follow-up before any mutation packet.
- **Claims:** Root + path hardening (`/` `/var` `/var/lib` root-owned, not group/world-writable); exclusive `0700` transaction dir; `0600` regular non-symlink files with hardlink count 1; noclobber lock file; flock on fd 9 with `/proc` inode match; atomic state via `mktemp` + `sync` + `mv -T`; reopen/status require exact identity + `schema:1` / `revision:0` / `recovery:{direction:forward,stage:created}`; artifact hashes bind staged `state.sh` + `observe.sh`. Identity is caller-supplied (not signed); header correctly defers Docker ownership proof to the host.
- **Prereqs:** Header requires root, `/proc`, `jq`, `flock`, GNU `stat`, `sha256sum`, `mktemp`, `sync`, `mv`. Aligns with a node-side private library, not a laptop learner shell.
- **Exercises / links:** N/A on the live site — good. Keep the ≤80-line lesson gated until the full transaction chain is independently accepted. Incomplete-state refuse + preserved baseline clusters in the PR body are operator evidence, not learner labs.
- **Tests:** In-repo unittest covers explicit identity schema (18 accept/refuse cases under a conditional caller) and source-time caller/option/descriptor preservation (fd 9 sentinel untouched). Matches the “inert when sourced” bar. Contention, wrong-identity, altered-state, unsafe perms, symlink, changed-artifact, and pre-rename sync-failure cases are claimed against owned v1.35.0 fixtures in the PR body — not yet reproduced as CI injects (same class of gap observe packet had outside CRI).

**Non-blocking follow-ups (not required to land this head)**
1. `cka_cert_state_init` sets `umask 077` for the lock create, but not before `mktemp` for `state.*`. Permissive caller umask fail-closes at the `600` check and leaves an incomplete transaction dir that blocks re-init — safe, but asymmetric with the lock path; set `umask 077` (or `mktemp -m 0600` where available) before the temp for clarity.
2. Measured packet 2 still expects verified backup custody + manifest/pair ownership checks. Land those (or an explicit split amendment on #2478) before integrating stop-consumer / renew / rollback.
3. Prefer at least one injected negative for lock contention and incomplete-state reopen refuse in CI so regressions do not depend solely on disposable fixture receipts.

CI was still settling at review time; this CF is exact-head content/safety review, not a green-checks substitute. No learner-facing publish yet.

PR #2515, opened 2026-09-10
feat(quality): consume markdown links via canonical resolver (#2346)

krisztiankoos:
Adds a markdown-link consumer script that extracts internal markdown hrefs, resolves them against a given manifest, and emits a revision-bound JSON report. Includes `node --test` fixtures for missing, sibling, fallback, UK-as-source, and fenced-code omission cases. (actual +274 lines).

### DIFF
--- a/scripts/quality/markdown-link-consumer.mjs
+++ b/scripts/quality/markdown-link-consumer.mjs
@@ -1,0 +1,112 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { parseArgs } from 'node:util';
+import { resolveRoute, UnsupportedRoute } from './route-resolver.mjs';
+import { buildDocumentRouteSet } from './document-route-set.mjs';
+
+const SOURCE_PREFIX = 'src/content/docs/';
+
+export function extractLinks(content) {
+  const links = [];
+  const lines = content.split('\n');
+  const linkRegex = /\[(?:[^\]\\]|\\.)*\]\(\s*([^)\s]+)(?:\s+"[^"]*")?\s*\)/g;
+  for (let i = 0; i < lines.length; i++) {
+    let match;
+    while ((match = linkRegex.exec(lines[i])) !== null) {
+      links.push({ href: match[1], line: i + 1 });
+    }
+  }
+  return links;
+}
+
+export function consumeMarkdownLinks(manifest, docsRoot) {
+  const routeSet = buildDocumentRouteSet(manifest);
+  const fallbackPaths = new Set(routeSet.fallbacks.map(f => f.pathname));
+  
+  const files = [];
+  function walk(dir) {
+    if (!fs.existsSync(dir)) return;
+    for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
+      const fullPath = path.join(dir, entry.name);
+      if (entry.isDirectory()) walk(fullPath);
+      else if (entry.isFile() && /\.(md|mdx)$/i.test(entry.name)) files.push(fullPath);
+    }
+  }
+  walk(docsRoot);
+  
+  const report = [];
+  
+  for (const file of files) {
+    const relativeToDocs = path.relative(docsRoot, file).split(path.sep).join('/');
+    const source = SOURCE_PREFIX + relativeToDocs;
+    
+    const primary = routeSet.primaryBySource.get(source);
+    if (!primary) continue;
+    
+    const content = fs.readFileSync(file, 'utf8');
+    const links = extractLinks(content);
+    
+    for (const { href, line } of links) {
+      let target = null;
+      let disposition = '';
+      
+      if (href.startsWith('#')) {
+        disposition = 'fragment-only';
+        target = new URL(href, primary.url).href;
+      } else {
+        try {
+          const resolved = resolveRoute(href, primary.url, routeSet.origin, routeSet.targetPaths);
+          target = resolved.url;
+          if (resolved.kind === 'external-http' || resolved.kind === 'mailto') {
+            disposition = 'external';
+          } else if (resolved.routeExists) {
+            disposition = fallbackPaths.has(resolved.path) ? 'present-fallback' : 'present';
+          } else {
+            disposition = 'missing';
+          }
+        } catch (error) {
+          if (error instanceof UnsupportedRoute) {
+            disposition = 'unsupported';
+          } else {
+            throw error;
+          }
+        }
+      }
+      
+      report.push({
+        source,
+        line,
+        href,
+        target,
+        disposition
+      });
+    }
+  }
+  return report;
+}
+
+if (import.meta.url === `file://${process.argv[1]}`) {
+  const { values, positionals } = parseArgs({
+    options: {
+      output: { type: 'string', default: '.agent/markdown-links.json' },
+    },
+    allowPositionals: true,
+  });
+  
+  const manifestPath = positionals[0];
+  if (!manifestPath) {
+    console.error('Usage: node markdown-link-consumer.mjs <manifest.json> [--output path]');
+    process.exit(1);
+  }
+  
+  const manifestRaw = fs.readFileSync(manifestPath, 'utf8');
+  const manifest = JSON.parse(manifestRaw);
+  
+  const docsRoot = path.resolve(process.cwd(), 'src/content/docs');
+  
+  const report = consumeMarkdownLinks(manifest, docsRoot);
+  
+  fs.mkdirSync(path.dirname(values.output), { recursive: true });
+  fs.writeFileSync(values.output, JSON.stringify(report, null, 2), 'utf8');
+  console.log(`Wrote ${report.length} links to ${values.output}`);
+}
--- a/scripts/quality/markdown-link-consumer.mjs
+++ b/scripts/quality/markdown-link-consumer.mjs
@@ -10,9 +10,18 @@ export function extractLinks(content) {
   const links = [];
   const lines = content.split('\n');
   const linkRegex = /\[(?:[^\]\\]|\\.)*\]\(\s*([^)\s]+)(?:\s+"[^"]*")?\s*\)/g;
+  let inFencedBlock = false;
   for (let i = 0; i < lines.length; i++) {
+    const line = lines[i];
+    if (line.trim().startsWith('```')) {
+      inFencedBlock = !inFencedBlock;
+      continue;
+    }
+    if (inFencedBlock) continue;
+    
+    let processedLine = line.replace(/`[^`]*`/g, '');
     let match;
-    while ((match = linkRegex.exec(lines[i])) !== null) {
+    while ((match = linkRegex.exec(processedLine)) !== null) {
       links.push({ href: match[1], line: i + 1 });
     }
   }
@@ -35,6 +44,7 @@ export function consumeMarkdownLinks(manifest, docsRoot) {
   walk(docsRoot);
   
   const report = [];
+  const memo = new Map();
   
   for (const file of files) {
     const relativeToDocs = path.relative(docsRoot, file).split(path.sep).join('/');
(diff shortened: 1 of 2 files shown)


krisztiankoos:
CF:CHANGES on head `afbae29989694082a23450d1ed8d8d0996189fa7` (quality consumer slice for #2346; no curriculum/book pages).

**What holds**
- Correctly layers on `resolveRoute` + `buildDocumentRouteSet` with browser URL semantics; the `./module-b/` false-pass resolves under the current directory and is reported `missing`, while `../module-b/` is `present` — matches the #2346 regression the issue asked for.
- UK fallback, index, explicit slug, root-relative, fragment-only, and unsupported (`ftp:`) fixtures are covered; `node --test` on the new file passes on this head.
- `SOURCE_PREFIX` matches `document-source-validation.mjs`; 2 files / ~232 lines is a coherent incremental consumer (gate wiring still open on #2346, same as prior producer/resolver slices).

**Needed before PASS**
1. **CLI entry never runs under a relative `node` path.** `import.meta.url === \`file://${process.argv[1]}\`` fails when argv is `scripts/quality/markdown-link-consumer.mjs` (naive `file://scripts/...` ≠ absolute `file:///...`). Use `pathToFileURL(process.argv[1]).href` (or resolve + `fileURLToPath`) so the Usage path in the script actually emits the JSON report.
2. **Image markdown is treated as links.** `extractLinks("diagram")` yields `/assets/x.png`, so a full-tree run will flood `missing` with static assets that are not document routes. Skip `` (or classify non-document targets explicitly) before this report is useful for the EN/UK link gate.

Not blocking this slice: wiring into `npm run check` / link-check.yml, and posting live-site failure paths — still open on #2346; do not close the issue on merge of consumer-only.

krisztiankoos:
## Review: PR #2515 — feat(quality): consume markdown links via canonical resolver (#2346)
**Reviewer**: Claude Sonnet 4.6 (Anthropic — cross-family vs agy/Google author)
**Round**: R1
**HEAD**: `afbae29989694082a23450d1ed8d8d0996189fa7`
**Verdict**: NEEDS_CHANGES

---

### Findings (priority-ranked)

#### P1 (blocker)
_None._

#### P2 (should fix)

**1. `resolveRoute` re-parses the full manifest route set on every single link — O(links × routes) — confirmed ~0.76ms/call** (`markdown-link-consumer.mjs:58`)

```js
// line 58 — called once per href:
const resolved = resolveRoute(href, primary.url, routeSet.origin, routeSet.targetPaths);
```

Inside `route-resolver.mjs`, `manifestRoutes(knownPaths, origin)` iterates over the full `targetPaths` Set and validates every path, returning a brand-new Set each call. Benchmark: 10k calls over 2,000 routes takes 7.6 s (0.76 ms/call). The KubeDojo corpus is ~2,000 pages with an estimated 20–40k links → **15–30 s of manifest re-validation overhead** on every run of the consumer.

`route-resolver.mjs` is off-limits to modify, so the fix is in the consumer: pre-invoke the outer validation once (or restructure to minimise re-entry). One workable approach — call a single-route resolution probe at startup to validate origin/manifest, then do the per-link lookup inline using the already-validated `routeSet.targetPaths` Set for the `routeExists` check, and delegate the full `resolveRoute` only for external/unsupported branching. If that duplicates logic, a lightweight memoisation wrapper around `resolveRoute` keyed on `(href, sourceUrl)` would reduce redundant re-entries for repeated same-href links without touching `route-resolver.mjs`. Needs a concrete decision before merge — the current form will time out on the live corpus.

---

**2. Links inside fenced code blocks and inline code are extracted as live links — false positives confirmed** (`markdown-link-consumer.mjs:12`)

`extractLinks` applies the regex line-by-line with no fence-state tracking. A code example like:

```markdown
```bash
# example only
kubectl get [service](./services/)
```
```

or inline code `[link](./path/)` will be extracted and may emit a `missing` or `present-fallback` disposition that is never a real broken link. KubeDojo docs are heavy with bash and YAML snippets. Even a single false-positive `missing` would make the consumer useless as a quality gate.

Fix: add a lightweight fence-state boolean (toggle on `/^```/`, skip lines while fenced; skip lines that start with ` ` for inline code). Exact snippet in context:

```js
// in extractLinks(), before linkRegex.exec():
let inFence = false;
for (let i = 0; i < lines.length; i++) {
  if (/^```/.test(lines[i])) { inFence = !inFence; continue; }
  if (inFence) continue;
  // … existing exec loop (also skip hrefs inside backtick spans on the line)
```

---

**3. No test for UK file as a link SOURCE (only as a link target)** (`markdown-link-consumer.test.mjs:8`)

The PR spec says "EN+UK markdown hrefs" but the single test fixture only extracts links from `dir1/module-a.md` (an EN file). The UK file `uk/dir1/module-b.md` in the fixture exists on disk but contains `'uk b content'` — no links — and its manifest entry has `source: 'src/content/docs/dir1/module-b.md'` (isFallback) so `primaryBySource.get('src/content/docs/uk/dir1/module-b.md')` returns `undefined` and the file is silently skipped.

To confirm the consumer actually processes a translated UK file's links (the advertised "UK" coverage), add a fixture route with `source: 'src/content/docs/uk/dir1/module-a.md'` + `isFallback: false`, create that file with at least one link, and assert the link appears in the report.

---

**4. PR body claims "Fits within file budget constraint" but additions are 232 (budget was ≤200)**

The PR description says it fits within the file budget constraint; the diff is +232 lines. The body needs correcting. The overrun is content (real tests, not padding) — but the false claim should not stay in the merged PR description.

---

#### Nits

**N1. `mailto:` links receive `'external'` disposition** (`markdown-link-consumer.mjs:60`)

```js
if (resolved.kind === 'external-http' || resolved.kind === 'mailto') {
  disposition = 'external';
}
```

`resolveRoute` returns `kind: 'mailto'` specifically. Collapsing it into `'external'` loses the distinction downstream. Consider `'mailto'` as its own disposition (or at minimum add a test so the current behaviour is documented as intentional).

**N2. `findReport` helper in tests matches by link TEXT substring, not href** (`markdown-link-consumer.test.mjs:87`)

```js
const findReport = (text) => report.find(r => r.href === text || mdContent.split('\n')[r.line - 1].includes(text));
```

The fallback arm matches the VISUAL link text ("broken sibling") against the raw line. Two links on the same line with the same visible text would collide. Matching on `r.href` directly (as the primary arm already tries) would be safer; just change the call sites to pass hrefs like `'./module-b/'` instead of display names.

**N3. Single monolithic `test()` — first assertion failure hides the rest** (`markdown-link-consumer.test.mjs:8`)

Eight distinct scenarios in one `test()` block. Use `t.test('...', ...)` subtests so each scenario is independently reported.

**N4. `SOURCE_PREFIX` hardcoded to `'src/content/docs/'`** (`markdown-link-consumer.mjs:7`)

This is implicitly coupled to the assumption that the caller passes exactly `src/content/docs` as `docsRoot`. If the constant and the argument diverge the consumer silently produces an empty report. Worth a brief comment and/or deriving it from `docsRoot` relative to `process.cwd()`.

---

### Verified

- [x] Tests: `node --test scripts/quality/markdown-link-consumer.test.mjs` → 1 pass, 0 fail ✓
- [x] Imports reuse `route-resolver.mjs` — no forked URL logic (`markdown-link-consumer.mjs:4`) ✓
- [x] `check_site_health.py` / `check_route_coherence.py` — NOT touched ✓
- [x] Broken sibling (`./module-b/` from `module-a/`) flagged `missing` under trailingSlash-always ✓
- [x] `present-fallback` disposition on isFallback route ✓
- [x] `routeSet.origin` (string) accepted by `siteOrigin()` in route-resolver ✓
- [x] Performance benchmark run (P2 #1 above — **confirmed slow**, 0.76ms/call)
- [x] Code-block false-positive confirmed (P2 #2 above)
- [ ] UK SOURCE file link extraction — not covered by tests (P2 #3)

### Out of scope
- `route-resolver.mjs`, `document-route-set.mjs` — not changed, reviewed only as interfaces
- `check_site_health.py`, `check_route_coherence.py` — explicitly excluded
- CI workflow integration (not included in this PR)

krisztiankoos:
Claude R1 NEEDS_CHANGES on `afbae2998` (no P1). AGY fix pass in flight: fence skip, UK-as-source fixture, resolveRoute memoization, honest +232 budget in the PR body.

krisztiankoos:
Claude R1 NEEDS_CHANGES on `afbae2998` (no P1). Dispatching AGY fix pass: fence skip, UK-as-source fixture, resolveRoute memoization, honest +232 budget in the PR body.

krisztiankoos:
## Review: PR #2515 — feat(quality): consume markdown links via canonical resolver (#2346)
**Reviewer**: Claude Sonnet 4.6 (Anthropic family)
**Round**: R2 (exact-head `db28889bfaa08d8a98c77098ec7bf76087d50c95`)
**Verdict**: APPROVE_WITH_NITS

---

### R1 findings re-check (all four items)

#### ✅ Item 1 — Fenced-block links skipped; `lastIndex` safety
**FIXED.**

Test fixture at `test.mjs:84-86`:
```markdown
```
[fenced link](/should/not/exist)
```
```
Assertion at `test.mjs:137-138`:
```js
const fencedLink = report.find(r => r.href === '/should/not/exist');
assert.ok(!fencedLink, 'Fenced link should not appear in the report');
```
Passes ✓

**`lastIndex` concern is a non-issue.** The regex at `consumer.mjs:12` is created once with `/g`. It is used in a `while ((match = linkRegex.exec(processedLine)) !== null)` loop that runs to exhaustion — no early `break` exists. Per spec, `exec()` returning `null` resets `lastIndex` to 0. Verified empirically:

```
after exhaustion, lastIndex: 0
```

The `continue` statements on the outer `for` loop skip to the next line only after the `while` loop has already reset `lastIndex`. No skipped-href risk.

#### ✅ Item 2 — UK source fixture (`isFallback: false`) consumed
**FIXED.**

Manifest route at `test.mjs:63-71`:
```js
{
  id: 'uk/module-c.md',
  source: 'src/content/docs/uk/module-c.md',
  isFallback: false,
  locale: 'uk',
  ...
}
```
File created at `test.mjs:98` with an internal link. Assertion at `test.mjs:140-143`:
```js
const ukSourceInternal = report.find(r =>
  r.source === 'src/content/docs/uk/module-c.md' && r.href === '../../dir1/module-a/'
);
assert.ok(ukSourceInternal, 'UK Source link not found');
assert.equal(ukSourceInternal.disposition, 'present');
```
Passes ✓

#### ✅ Item 3 — `resolveRoute` memoized in consumer only; `route-resolver.mjs` untouched
**FIXED.**

Memo at `consumer.mjs:47`: `const memo = new Map();`
Memoization logic at `consumer.mjs:67-90` (key: `${href}::${primary.url}`).
`route-resolver.mjs` is not in this PR's diff. ✓

#### Nit — Stale line-count claim in PR body
**STILL PRESENT.**

PR body reads: `(actual +232 lines)`
Actual additions per `gh pr view 2515 --json additions`: **274**

The body predates the fix commit (`db28889bf`) which added the fenced-block and UK-source fixtures. The claim is stale by 42 lines. Low severity (PR body only, not code), but worth correcting for reviewers reading the history.

---

### Verification

```
node --test scripts/quality/markdown-link-consumer.test.mjs

✔ consumeMarkdownLinks handles various fixtures (2.547791ms)
ℹ tests 1
ℹ pass 1
ℹ fail 0
```

---

### Out of scope
- `document-route-set.mjs` and `route-resolver.mjs` (not touched by this PR; not reviewed)
- CI / workflow configuration
- Integration against the live manifest

krisztiankoos:
AGY fix landed at `db28889bf`. Local `node --test` green. Claude R2 dispatched on the new head; do not merge until exact-head APPROVE + CI on this SHA.

krisztiankoos:
CF:CHANGES on head `db28889bfaa08d8a98c77098ec7bf76087d50c95` (quality consumer for #2346; no curriculum/book pages).

**Landed since prior CF**
- Fenced-block skip + fixture: `[fenced link](/should/not/exist)` stays out of the report.
- UK-as-source (`uk/module-c.md`, `isFallback: false`) is consumed and asserts `present` for an EN-relative href.
- Per-`(href, primary.url)` memo around `resolveRoute` only; resolver untouched.
- #2346 `./module-b/` false-pass → `missing` vs `../module-b/` → `present` still holds.

**Still needed before PASS** (same as CF on `afbae2998`)
1. **CLI entry never runs under a relative `node` path.** `import.meta.url === \`file://${process.argv[1]}\`` still fails when argv is `scripts/quality/markdown-link-consumer.mjs` (naive `file://scripts/...` ≠ absolute `file:///...`). Use `pathToFileURL(resolve(process.argv[1])).href` (or `fileURLToPath` compare) so the Usage path actually writes the JSON report.
2. **Image markdown is still treated as links.** `extractLinks("diagram")` yields `/assets/x.png` (verified: the link regex matches inside ``). A full-tree run will flood `missing` with static assets that are not document routes. Skip `` (or classify non-document targets) before this report is gate-ready.

Nit (non-blocking): PR body still says `(actual +232 lines)` while this head is +274 additions.

Out of scope for PASS on this slice: wiring into `npm run check` / link-check.yml — still open on #2346; do not close the issue on consumer-only merge.

Issue #2479, opened 2026-09-09
feat(cka): validate isolated etcd snapshot restoration and rollback


krisztiankoos:
Parent #2280; full upgrade #2272. Follow-up to closed #2338. Source boundary card #2466 and containment #2468 removed unvalidated snapshot/restore mutation recipes; practical etcd restoration has not run.

Design and validate a separate dedicated disposable Kubernetes/etcd fixture before publishing executable restore practice. Own only the etcd recovery artifact and its eventual bounded module-5.3 lesson integration. Do not mutate shared/live infrastructure or the existing persistent fixture, delete unrelated artifacts, or replace existing services.

Start with a source-verified topology/tool compatibility contract and independently reviewed split. Default <=200 aggregate changed lines and <20 files per PR. If the complete safe transaction cannot fit, post its measured split before implementation; no incomplete recovery path may be presented as executable learner practice.

Required results:
- Explicit Kubernetes/etcd/tool versions, cluster/container ownership, unique private artifacts, collision refusal and retention of original data.
- Authenticated baseline, a marker captured by the snapshot, and a separately observed post-snapshot mutation.
- Reviewed backup/restore sequence with revision/compaction handling matched to the topology; bounded waits and interruption-safe recovery.
- Verify restored contents and fresh API/controller behavior, not merely snapshot readability or process startup. Preserve evidence of rollback to the original state and refuse cleanup when identity/recovery is unresolved.
- Actual positive, failure and interruption evidence on the owned fixture, with deterministic injected checks clearly separated. No fixed sleep is health proof, and fixture destruction is not rollback proof.
- Independent-family review before merge and before lesson publication. Preserve diagrams and the accepted scheduler exercise; align outcomes with observed results and verify build/render/copy behavior.

No etcd restoration has executed. Track practical certificate renewal separately. Whole-module, hosted-lab and Ukrainian acceptance remain outside this issue.

krisztiankoos:
## Overnight pickup after #2478 close

Starting measured split for isolated etcd snapshot restore (≤200 LOC/PR), mirroring cert fixture discipline:

1. **E0** — topology/tool pin + ownership contract (doc or tiny fixture stub; no restore yet)
2. **E1** — disposable fixture lifecycle (create/inspect/delete) if not reusable from cert fixture
3. **E2** — snapshot + marker + post-snapshot mutation observation
4. **E3** — restore + verify marker/API; rollback/interrupt labelled separately
[...]

PR #2562, opened 2026-09-13
fix(cka): use etcdctl --keys-only for marker baseline (#2479)

krisztiankoos:
## Summary
Live evidence failed: `etcdctl get` of a ConfigMap key returns protobuf bytes; `Fixture.run(..., text=True)` then raises UTF-8 decode errors. Use `--keys-only` for the existence check.

## Test plan
- [x] focused unittest
- [ ] CF APPROVE + merge
- [ ] re-run live `.agent/check-cka-etcd-restore.py`

### DIFF
--- a/scripts/cka_etcd_fixture.py
+++ b/scripts/cka_etcd_fixture.py
@@ -197,7 +197,8 @@ class EtcdFixture:
         status = json.loads(self._etcdctl("endpoint", "status", "-w", "json"))[0]["Status"]
         revision = status["header"]["revision"]
         etcd_key = f"/registry/configmaps/{marker['namespace']}/{marker['name']}"
-        if not self._etcdctl("get", etcd_key):
+        # Prefer keys-only: raw ConfigMap values are protobuf and break text=True capture.
+        if not self._etcdctl("get", etcd_key, "--keys-only"):
             raise RuntimeError("Marker key missing in etcd")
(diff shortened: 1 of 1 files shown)


krisztiankoos:
CF:CHANGES

**Situational:** Only open PR; linked #2479; no competing `codex/*` PRs.

**Diff review (`scripts/cka_etcd_fixture.py`):** The `--keys-only` switch on the marker `etcdctl get` is the right fix. `Fixture.run` always uses `text=True`, and raw ConfigMap values in etcd are protobuf, so a full `get` blows up on UTF-8 decode before the existence check can succeed. Existence-only is enough here: marker *content* is still checked via the API in `ensure_marker()` / `_marker_cm`. Narrow, learner-safe for the #2479 fixture path.

**Ask before PASS:**
1. Land a focused unit assert that the baseline `get` includes `--keys-only` (e.g. in `test_snapshot_observe_success` or a dedicated `etcd_baseline` test). The PR checklist marks focused unittest done, but `scripts/tests/test_cka_etcd_fixture.py` is unchanged and the current mock returns `"present"` for any non-save `etcdctl` shell, so it would still pass without the flag. That leaves the protobuf/`text=True` regression unguarded in CI.
2. Keep the unchecked live re-run of `.agent/check-cka-etcd-restore.py` as author follow-up (not a CF content block once the unit assert lands).

No implement from CF. Head `10c7d5a`.

PR #2563, opened 2026-09-13
fix(cka): stage etcd snapshots under /var/lib for docker cp (#2479)

krisztiankoos:
## Summary
Live evidence: `docker cp` cannot see files under kind node `/tmp` (tmpfs). Stage snapshot save/restore copies under `/var/lib/etcd-fixture-transaction/` instead.

## Test plan
- [x] focused unittest
- [ ] CF + merge
- [ ] re-run live evidence harness

### DIFF
--- a/scripts/cka_etcd_fixture.py
+++ b/scripts/cka_etcd_fixture.py
@@ -214,7 +214,9 @@ class EtcdFixture:
         snap_path = self.inner.directory / SNAPSHOT_FILE
         if self.inner.state.get("snapshot") or snap_path.exists():
             raise RuntimeError("Refusing to overwrite existing snapshot")
-        node_snap = "/tmp/etcd-snapshot-" + uuid.uuid4().hex + ".db"
+        self._node_sh(f"mkdir -p {RESTORE_TXN}")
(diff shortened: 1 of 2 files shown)


krisztiankoos:
CF:PASS

Surgical fix for kind `docker cp` vs node `/tmp` (tmpfs / not on the rootfs docker sees). Staging under existing `RESTORE_TXN` (`/var/lib/etcd-fixture-transaction`) matches the yaml staging pattern and correctly covers both directions: snapshot save (`cp` out) and offline restore (`cp` in).

- Claims: root cause and path choice look right; no leftover `/tmp` snap paths in those two sites.
- Tests: unittest mock updated for the new `mkdir -p` on save.
- Scope: scripts only; no lesson text drift. Linked #2479 E3 follow-up; no competing open `codex/*` / content PRs.
- Non-blocking: live evidence harness re-run still open on the PR checklist — expected next step after merge, not a block on this path fix. Node-side snap files under `RESTORE_TXN` still left in place (same as prior `/tmp` temps); fine for a disposable fixture.

Head `b4d38d49` · ready for merge once CI settles.

Issue #3, opened 2025-12-02
Part 2: Workloads & Scheduling (15%)


krisztiankoos:
## Overview
Core workload management - deploying, scheduling, and scaling applications.

## Modules (Updated for CKA 2025)

- [ ] **2.1 Pods** - What actually happens when you create a pod (API → scheduler → kubelet → runtime)
- [ ] **2.2 Deployments, ReplicaSets, DaemonSets** - Why each exists, when to use which
- [ ] **2.3 Jobs & CronJobs** - Batch processing, completion modes, failure handling
- [ ] **2.4 Scheduling** - nodeSelector, affinity/anti-affinity, taints, tolerations
- [ ] **2.5 Resource Requests & Limits** - QoS classes, OOM behavior, resource quotas
- [ ] **2.6 Pod Admission & Policies** - Pod Security Admission, security contexts 🆕
- [ ] **2.7 Workload Autoscaling** - HPA, VPA basics, scaling behavior 🆕

## Exam Weight
**15%** - Approximately 2-3 questions

## Key Skills (2025)
- Create and manage deployments
- Configure pod scheduling constraints
- Set appropriate resource limits
- Use taints and tolerations
- Configure Pod Security Admission
- Set up Horizontal Pod Autoscaler

## What Changed in 2025
| Added | Notes |
|-------|-------|
| Pod Admission controls | Security contexts, PSA |
| Workload autoscaling | HPA focus |
| Self-healing emphasis | Resilience patterns |

## Common Gotchas
- Affinity vs anti-affinity syntax confusion
- Forgetting tolerations when taints exist
- Resource limits vs requests behavior
- HPA requires metrics-server
- Pod Security Admission modes (enforce, audit, warn)

---
⚠️ **Curriculum Version**: CKA 2025 (effective Feb 18, 2025)
📋 **Source**: [CNCF Curriculum](https://github.com/cncf/curriculum)

PR #2756, opened 2026-09-18
feat(agent): TypeSafe/Jev client, epic triage, cookbook inventory

krisztiankoos:
## Summary
- `scripts/typesafe_client.py` — System One HTTP client, receipts, `top_prob` / citation helpers
- `scripts/jev_epic_triage.py` — batched pre-dispatch triage with family exclusion + parallel policy
- `agents_extensions/shared/skills/typesafe-ai/` — SKILL, BEST_PRACTICES, COOKBOOKS (18-cookbook adopt/park)

## Test plan
- [ ] `.venv/bin/ruff check scripts/typesafe_client.py scripts/jev_epic_triage.py`
- [ ] Helper smoke without live API
- [ ] Confirm no curriculum `src/content` changes


krisztiankoos:
## Review: PR #2756 — feat(agent): TypeSafe/Jev client, epic triage, cookbook inventory
**Reviewer**: grok-4.6 (xAI Grok) — CF ≠ Cursor author
**Round**: R1
**Head**: `73086d68c1face0e8f6f9abe83d89556b5048f8c`
**Verdict**: NEEDS_CHANGES

Cross-family R1 of `scripts/typesafe_client.py`, `scripts/jev_epic_triage.py`, and `agents_extensions/shared/skills/typesafe-ai/` (`.claude/skills/typesafe-ai/` is a matching deploy mirror). Scope checks requested by the dispatch: **no secrets in tree**, **thresholds in code**, **no `src/content`**.

### Findings (priority-ranked)

#### P1 (blocker — must fix before merge)

1. `scripts/jev_epic_triage.py:372` — `_repick_seat` can return an **excluded** family. The PR advertises a code-side CF≠author / CF≠auditor post-filter (`Friction fixes` item 3). When every `SEAT_ORDER` seat is in `exclude`, the hardcoded `"cursor"` fallback still wins — including when `cursor` itself is excluded.

FINDING: `_repick_seat` returns excluded family `cursor` when SEAT_ORDER is fully excluded
FILE:LINE: scripts/jev_epic_triage.py:367
CURRENT CODE:
    if preferred and preferred not in exclude and preferred != "none":
        return preferred
    for seat in SEAT_ORDER:
        if seat not in exclude:
            return seat
    return "cursor"
WHY: Two CF-missing PRs with four distinct author/auditor families is a normal conveyor shape (e.g. cursor+agy and kimi+grok). `exclude` then equals `{kimi, agy, grok, cursor}`. The filter claims “family exclusion” while recommending the excluded author family as CF. Exercised locally:

```text
_repick_seat('kimi', set(SEAT_ORDER)) → 'cursor'
recommend(... exclude=all SEAT_ORDER ...) → cf_agent='cursor'
  overrides=['cf_agent kimi→cursor (family exclusion)']
  exclude_families_for_cf=['agy','cursor','grok','kimi']
```

The override string says exclusion was applied; the value violates it. `advisory_only` does not make a lying post-filter acceptable — leads will follow `cf_agent`.
FIX: Return `"none"` (or raise/escalate) when no eligible seat remains. Never return a member of `exclude`. If `adj_cf == "none"`, append `escalate` like `no eligible CF seat after family exclusion` and do not recommend dispatch_cf onto an excluded family.

#### P2 (should fix)

1. `scripts/typesafe_client.py:213` — `choice_is_uncertain` **fail-open** when a Choice has a label but neither `probabilities` nor `confidence`. Cookbook / COOKBOOKS.md say auto-act iff `max(probabilities) ≥ 0.60`. Missing peak should be uncertain.

FINDING: choice_is_uncertain returns False (auto-act) when probabilities and confidence are both missing
FILE:LINE: scripts/typesafe_client.py:210
CURRENT CODE:
    if not ans:
        return True
    conf = ans.get("confidence")
    if conf is None:
        return False
    return float(conf) < min_top
WHY: Exercised: `choice_is_uncertain({'type':'choice','choice':'kimi'}) → False`. `recommend()` then treats `next_action` / `cf_agent` as certain (`next_action_uncertain=false`, empty `escalate`). A truncated/malformed System One payload auto-acts instead of escalating.
FIX: `if conf is None: return True` (fail closed). Optionally treat missing `probabilities` as uncertain even when `confidence` is present, matching the cookbook’s `max(probabilities)` gate.

2. `scripts/typesafe_client.py:137` / `scripts/jev_epic_triage.py:518` — receipts and `decisions.jsonl` drop Choice `probabilities`. Threshold tuning is supposed to come from experiment receipts (`MIN_CHOICE_TOP_PROB`, `AMBIGUOUS_GAP`), but `_write_receipt` keeps only `choice` + `confidence`. After a live run you cannot recompute `top_prob` or top-two gap from disk.

FINDING: compact Choice receipts omit probabilities needed to tune top_prob gates
FILE:LINE: scripts/typesafe_client.py:137
CURRENT CODE:
        elif t == "choice":
            compact_answers[qid] = {
                "type": "choice",
                "choice": ans.get("choice"),
                "confidence": ans.get("confidence"),
            }
WHY: I wrote a dummy receipt via `_write_receipt`; the JSONL row had `choice`/`confidence` and no `probabilities`. `confidence` is not a substitute for `max(probabilities)` or the 0.15 gap test.
FIX: Persist `probabilities` (and noul already persists). Same for `append_decision` Choice branch.

3. `scripts/jev_epic_triage.py:159` — `_cf_status` only inspects `comments[-5:]`. A real CF verdict older than five later comments becomes `missing` and the tool will recommend `dispatch_cf` again. `NEEDS_CHANGES` vs `APPROVE` order inside that window is fail-closed (good); the window itself is not.

FIX: Scan all comment bodies (or the newest verdict, not the newest five comments). Keep NEEDS_CHANGES-before-PASS.

4. **No tests.** `recommend`, `_repick_seat`, `choice_is_uncertain`, `check_claim_against_excerpt` (fabricated-quote short-circuit), and `_ci_status` are pure and currently untested. `rg` over `scripts/tests` / `test*.py` found zero references. The P1/P2 above were caught with a 40-line ad-hoc driver, not CI. Add `scripts/tests/test_jev_epic_triage.py` + `test_typesafe_client.py` covering at least: all-exclude `_repick_seat`, parallel_wave + noul `< 0.35` → serialize/`dispatch_cf`, fabricated quote does not call HTTP, `choice_is_uncertain` missing peak.

5. `agents_extensions/shared/skills/typesafe-ai/BEST_PRACTICES.md:135` — Playbook A lists questions `path_disjoint` Noul and `difficulty` Score. `default_questions()` in `jev_epic_triage.py` has `next_action`, `cf_agent`, `author_agent`, `parallel_ok`, `jev_confidence_note` — no `path_disjoint` / `difficulty`. Docs and code will drift the first time someone copies the playbook.

#### Nits (optional)

1. `scripts/typesafe_client.py:31` — `KEY_CANDIDATES` includes CWD `Path(".secrets") / "typesafe-ai.key"` while BEST_PRACTICES says host path is `~/.secrets/` **(not repo `.secrets/`)**. `*.key` is gitignored, so this is not a commit leak, but it contradicts the skill. Prefer only `Path.home() / ".secrets" / "typesafe-ai.key"` plus `TYPESAFE_API_KEY`.

2. `scripts/jev_epic_triage.py:141` — `_ci_status` treats `{status: COMPLETED, conclusion: ""}` as green (`completed-no-conclusion → (True, [])`). Sibling `scripts/lib/pr_merge.py` does similar (`status == "COMPLETED"`). Residual false-green risk for `merge_ready_hint`; acceptable only because merge is not this script’s job.

3. `scripts/typesafe_client.py:111` — `system_one` raises `SystemExit` on HTTP errors. Fine for CLI; hostile if this stays the library import for later callers. Prefer a small `TypeSafeError` and let `main()` exit.

4. `scripts/jev_epic_triage.py:207` — state sent to Jev truncates `headRefOid` to 7 chars. Advisory only; exact-head CI remains a lead/merge concern.

5. `_enrich_pr` and `_cf_status` each hit the comments API (N+1). Last-comment author inference + last-5 CF scan could share one fetch.

### Scope checks (requested)

| Check | Result |
| --- | --- |
| Secrets in tree | Pass. No hardcoded key. Receipts log `model` + usage only. Dummy `_write_receipt` with `api_key='SHOULD_NOT_APPEAR'` did not persist it. Load path is env / `~/.secrets/typesafe-ai.key`. HTTP `Authorization` is not printed. |
| Thresholds in code | Pass. `MIN_CHOICE_TOP_PROB=0.60`, `CITATION_AUTO_ACCEPT=0.80`, `AMBIGUOUS_GAP=0.15`, `PARALLEL_NO=0.35`, `PARALLEL_YES=0.65` are Python constants; skill text points at them. |
| `src/content` | Pass. `git diff --name-only origin/main...HEAD` is the 8 skill/script files only. |

### Verified

- [x] Linter: `.venv/bin/ruff check scripts/typesafe_client.py scripts/jev_epic_triage.py` → `All checks passed!`
- [x] `bash agents_extensions/deploy.sh --check --target claude` → `'claude' extensions in sync`
- [x] Skill copies: `diff -q` shared vs `.claude/skills/typesafe-ai/{SKILL,BEST_PRACTICES,COOKBOOKS}.md` identical
- [x] Tests: **none in repo** for these modules. Ad-hoc `.venv/bin/python` driver exercised `_repick_seat`, `recommend`, `choice_is_uncertain`, `_ci_status`, `_infer_*_family`, fabricated-quote short-circuit, receipt redaction (output quoted above)
- [x] Density gates: N/A (not curriculum modules)
- [x] Citations: cookbook / `docs.typesafe.ai` URLs **UNVERIFIABLE** from this seat (no live docs fetch used). Not asserted real or fake.
- [x] Lab snippets: N/A
- [x] `.agent/` and `*.key` are gitignored; receipts cannot land in git by default
- [ ] Live TypeSafe HTTP call: **not executed** (no need to spend quota to review client shape)

### Out of scope

- Live System One API schema / `jev-latest` resolution / cookbook page bodies (UNVERIFIABLE here)
- Dispatch/merge/CF posting behavior (this PR’s scripts do not dispatch, merge, or post CF — docstring holds)
- Curriculum modules, GitHub workflows, `src/content`
- Whether Cursor vs Grok family pairing is the standing Decision Card C default (this comment is the requested Grok CF on a Cursor-authored PR)

**R2 expectation:** fail-closed `_repick_seat`, fail-closed `choice_is_uncertain`, persist `probabilities` on receipts, and a small pytest module for those cases. After that this is APPROVE_WITH_NITS at worst.

krisztiankoos:
## Cross-Family Review — PR #2756

**Reviewer family:** Claude (author family: Cursor/epic-driver — CF ≠ author ✅)
**Exact head SHA reviewed:** `73086d68c1face0e8f6f9abe83d89556b5048f8c`
**CI on that exact SHA:** 11/11 `success`, no drift between review start and finish ✅

### VERDICT: NEEDS_CHANGES

---

### Gate checks (all pass)

| Gate | Result |
|---|---|
| Secrets in diff | Clean — no keys, tokens, or PEM material. Key is read from env/`~/.secrets` only, never logged; `_write_receipt` emits `model`/`usage` only |
| `.venv/bin/ruff check` | **Clean** on both new files (the `EXE001` I first saw was my own `/tmp` copy dropping the exec bit; git has both at mode `100755`) |
| `ruff format --check` | Clean |
| `src/content/` touched | **0 files** ✅ |
| Skill mirror parity | `.claude/skills/typesafe-ai/*` is byte-identical to `agents_extensions/shared/skills/typesafe-ai/*` (SHA-256 match on all three), consistent with "Extensions in sync" passing |
| Scope / budget | 8 files, +1696/−0, one concern, no generated artifacts ✅ |

The documentation is genuinely good. `BEST_PRACTICES.md` §5 and the COOKBOOKS adopt/park split are the kind of durable fleet reference that survives a model swap, and the repeated "Jev advises; CF ≠ author + exact-head CI remain merge authority" framing is the right guardrail to state up front.

---

## P1 findings

### P1-1 — `_repick_seat` fails open and silently recommends an *excluded* CF family, while printing an override that claims exclusion was applied

This is the headline feature of the file (docstring friction fix #3: *"CF/author post-filter excludes author_family + auditor_family"*). It does not hold.

`SEAT_ORDER` has exactly four entries, so once all four are excluded the fallback returns `"cursor"` unconditionally — without consulting `exclude`:

```python
def _repick_seat(preferred: str | None, exclude: set[str]) -> str:
    if preferred and preferred not in exclude and preferred != "none":
        return preferred
    for seat in SEAT_ORDER:
        if seat not in exclude:
            return seat
    return "cursor"   # <-- returned even when "cursor" is in `exclude`
```

Reproducer against the head SHA (two open PRs, four families excluded — an ordinary busy-epic state):

```
RECOMMENDED cf_agent  = cursor  (excluded set: ['agy', 'cursor', 'grok', 'kimi'])
escalate              = []
overrides             = ['cf_agent kimi→cursor (family exclusion)']

>>> cf_agent in excluded set? True
```

Three things go wrong at once: the recommended CF equals an author family, `escalate` is **empty** so nothing signals the violation, and the `overrides` string asserts `(family exclusion)` — affirmatively telling the lead that the filter ran clean when it just failed. A tool that is wrong is recoverable; a tool that is wrong *and* reports success is not, because it removes the lead's reason to double-check.

Suggested fix — make the fallback refuse rather than guess:

```python
def _repick_seat(preferred: str | None, exclude: set[str]) -> str | None:
    if preferred and preferred not in exclude and preferred != "none":
        return preferred
    for seat in SEAT_ORDER:
        if seat not in exclude:
            return seat
    return None  # caller must escalate: no clean seat available
```

and in `recommend()`, append to `escalate` (not `overrides`) when it returns `None`.

Related, worth fixing in the same pass: `SEAT_ORDER` omits `claude`, `codex`, and `deepseek`, all of which are valid CF families here — I am a `claude` seat writing this review. `_repick_seat("claude", {...all four...})` correctly returns `claude` when Jev names it, but the fallback scan can never reach those families on its own, which is what makes the dead-end reachable in the first place. Widening `SEAT_ORDER` shrinks the failure window even after the `None` fix lands.

### P1-2 — `merge_ready_hint` can be `True` on a PR that is not merge-ready (two independent fail-opens, both biased toward "merge")

`merge_ready_hint` feeds the `merge_green_cf_done` recommendation, so both of these push the conveyor toward merging.

**(a) `_ci_status` counts unknown states as green.** Two real rollup shapes fall through every branch and are never appended to `pending`:

```
COMPLETED w/ no conclusion : (True, [])
legacy StatusContext(FAIL) : (True, [])
```

The second matters most: `statusCheckRollup` mixes `CheckRun` and `StatusContext` entries, and `StatusContext` carries `state`, not `status`/`conclusion`. A *failing* external commit status is therefore invisible and the PR reads green. Suggested fix: treat an entry matching no known-good shape as pending rather than passing, and read `state` for `StatusContext`.

**(b) `_cf_status` is not head-pinned.** It scans the last five issue comments for a verdict string with no reference to the commit the verdict was written against. A CF that passed three pushes ago reads as `pass` today. Given `exact_head_ci_required_to_merge: True` is asserted in the very `policy` block this function's output sits beside, the check should compare the CF comment's `created_at` against the head commit's timestamp (or parse the head SHA the CF pinned) before returning `pass`.

---

## P2 findings

- **`choice_is_uncertain` fails open on a malformed answer.** With neither `probabilities` nor `confidence` present it returns `False` — i.e. "certain, go ahead": `choice_is_uncertain({'choice': 'a'})` → `False`. For a helper whose whole job is the auto-act gate, absent evidence should read as uncertain, not as confidence. Change the `conf is None` branch to `return True`.
- **Repo-relative `.secrets` fallback contradicts the skill doc.** `KEY_CANDIDATES` includes `Path(".secrets") / "typesafe-ai.key"`, which resolves against the *current working directory*, not the repo root — so the key loaded depends on where the agent happened to `cd`. `BEST_PRACTICES.md` explicitly says "Host path is `~/.secrets/` (not repo `.secrets/`)". Not a commit-leak risk (`.gitignore:59 *.key` covers it — I confirmed with `git check-ignore`), but the code and its own documentation disagree. Drop the entry, or anchor it to `repo_root()`.
- **`_cf_status` regex false-positives on prose.** `NEEDS[_\s-]?CHANGES` is matched against uppercased comment bodies, so a comment reading "VERDICT: APPROVE (not NEEDS_CHANGES)" classifies as `needs_changes`. Anchor to a line-start verdict marker.
- **Two `gh api` calls per PR hit the same endpoint.** `_enrich_pr` fetches `.[-1].body` and `_cf_status` fetches `[.[].body]` from `issues/{n}/comments` — at `--limit 12` that is 24 round-trips where 12 would do. Fetch once and pass the list down.
- **Dead conditional in the `cf_agent` return.** Both arms of `adj_cf if next_choice in (...) else (adj_cf if raw_cf != "none" else "none")` evaluate to `adj_cf` except when `raw_cf == "none"`, so the outer `next_choice` test has no effect. Collapse it.
- **PR body test plan is unchecked.** All three boxes are `- [ ]`. I verified all three independently and they pass, but per the pre-submit checklist they should be ticked by the author before submit.

---

### What would flip this to APPROVE

P1-1 (fallback returns `None` + escalates, and the override string stops claiming exclusion ran) and P1-2 (unknown CI states read as pending; `_cf_status` head-pinned). The P2s are worth folding into the same push but none of them alone would block.

Nothing here regresses existing behavior — both scripts are net-new with no call sites — so this is a "fix before the fleet relies on it" review, not a "this broke something" review. The reason I am not waving it through on that basis is that the defect sits precisely on the CF ≠ author invariant the tool exists to protect, and it reports success while violating it.

*Cross-family review per `docs/review-protocol.md`. Posted as a comment rather than `gh pr review` because author and reviewer share a GitHub identity.*

krisztiankoos:
## Cross-Family Review (R2) — PR #2756

**Reviewer family:** Claude (author family: Cursor — CF ≠ author ✅)
**Round:** R2 (re-review after NEEDS_CHANGES fix-pass)
**Exact head SHA reviewed:** `6af48f0af9c72b1f44687bcf253216b50b480892`
**Fix commit:** `6af48f0` on top of `ddd243e`
**CI on that exact SHA:** 11/11 `success` (`gh api .../commits/6af48f0af9c72b1f44687bcf253216b50b480892/check-runs` — CodeQL, Astro build, Extensions in sync, Unit tests, Incident dedup gate, arXiv citation check, Site health link check, detect, Analyze × 3). No drift between review start and finish ✅

### VERDICT: APPROVE

---

## R1 findings — disposition

Every finding was re-exercised against the head SHA with a driver loading both modules from the worktree at `6af48f0`, not from the diff.

| R1 finding | Status |
|---|---|
| **P1-1** `_repick_seat` fails open, returns an excluded family, `overrides` claims exclusion ran | **Fixed** ✅ |
| **P1-2a** `_ci_status` counts unknown states as green | **Fixed** ✅ |
| **P1-2b** `_cf_status` not head-pinned | **Fixed** ✅ |
| **P2** `choice_is_uncertain` fails open on missing confidence | **Fixed** ✅ |
| **P2** repo-relative `.secrets` fallback contradicts BEST_PRACTICES | **Fixed** ✅ |
| **P2** `_cf_status` regex false-positives on prose | **Fixed** ✅ |
| **P2** dead conditional in the `cf_agent` return | **Fixed** ✅ (collapsed into `cf_out`) |
| **P2** two `gh api` calls per PR hit the same endpoint | Not addressed — carry-forward |
| **Nit** PR body test plan unchecked | Not addressed |
| *(grok R1)* **P2** receipts drop Choice `probabilities` | Not addressed — carry-forward |
| *(grok R1)* **P2** no tests | Not addressed — carry-forward, see below |
| *(grok R1)* **P2** BEST_PRACTICES Playbook A lists `path_disjoint`/`difficulty`, absent from `default_questions` | Not addressed — carry-forward |

### P1-1 verified fixed

`SEAT_ORDER` widened from 4 to 7 (`claude`, `codex`, `deepseek` added), `_repick_seat` returns `str | None`, and the caller escalates. Re-running my R1 reproducer — two open PRs, every family excluded:

```
_repick_seat('kimi', ALL_SEATS) -> None
_repick_seat(None, ALL_SEATS)   -> None

recommend(...) with all 7 families excluded:
  cf_agent   = None
  escalate   = ["no CF seat outside excluded families ['agy','claude','codex','cursor','deepseek','grok','kimi']", ...]
  overrides  = ["cf_agent no clean seat (excluded=[...]); escalate"]
  cf_agent in exclude? False
```

All three R1 failure modes are gone: the recommendation is `null` rather than an excluded family, `escalate` is populated with the accurate reason, and the `overrides` string no longer asserts `(family exclusion)` succeeded. The added brief line — *"If cf_agent is null, do not dispatch — escalate"* — closes the loop on the reader side. Partial-exclusion behaviour still works: `_repick_seat('kimi', {kimi,agy,grok,cursor})` → `'claude'`, and the normal two-family-excluded case returns `cf_agent='kimi'`, `author_agent='agy'`, `escalate=[]`.

### P1-2a verified fixed

Both shapes I named now read as not-green:

```
all success             -> (True,  [])
COMPLETED no conclusion -> (False, ['a:unknown'])
StatusContext FAILURE   -> (False, ['legacy:unknown'])
in progress             -> (False, ['a'])
empty rollup            -> (False, ['unknown'])
```

The catch-all `pending.append(f"{name}:unknown")` is the right shape — anything not matching a known-good state is pending.

### P1-2b verified fixed

`_cf_status(pr, head_oid)` now requires the comment to carry the full or short SHA, and scans **all** comments rather than the last five (this also closes grok's R1 P2-3):

```
exact-head APPROVE                  -> pass
exact-head NEEDS_CHANGES            -> needs_changes
stale head APPROVE                  -> missing        <- was `pass` at R1
prose "APPROVE (not NEEDS_CHANGES)" -> pass           <- was `needs_changes` at R1
old CF buried under 9 newer comments-> pass           <- was `missing` at R1
```

---

## P2 findings at this head (none blocking; all recommended before the fleet routes on `cf_status`)

### P2-1 — the verdict regex does not recognize the format real CF comments in this repo actually use

`^\s*(\*\*)?verdict(\*\*)?\s*:` allows a bold wrapper but not a Markdown heading prefix. Both R1 comments on this PR were run through the fixed function:

```
PR #2756: 2 comments
  [0] verdict line: ['**Verdict**: NEEDS_CHANGES']    (grok R1)
  [1] verdict line: ['### VERDICT: NEEDS_CHANGES']    (claude R1)
  _cf_status(pinned to R1 head 73086d6) -> present_unknown     <- expected needs_changes

PR #2758: 1 comment
  [0] verdict line: ['### VERDICT: NEEDS_CHANGES']
  _cf_status(pinned to R1 head 625db97) -> present_unknown     <- expected needs_changes
```

Two things combine. The `###`-prefixed line matches neither verdict branch, so it falls through to the `present_unknown` catch-all; and because the loop `return`s on the first matching comment (newest first), the correctly-formatted `**Verdict**: NEEDS_CHANGES` one comment older is never reached. Net effect: on the two PRs this tool exists to triage, the exact-head CF gate cannot read either verdict.

This lands fail-closed — `present_unknown` is not `pass`, so `merge_ready_hint` stays `False` and nothing gets waved through. The cost is that the gate is effectively inert and will keep recommending `dispatch_cf` for PRs that already have CF. Fix is one character class: `^\s*(#{1,6}\s*)?(\*\*)?verdict(\*\*)?\s*:`.

### P2-2 — newest-comment-wins drops the NEEDS_CHANGES-beats-APPROVE ordering that R1 called out as correct

The old implementation joined bodies and tested `needs_changes` first, so a NEEDS_CHANGES anywhere in the window won. The new loop returns on the newest matching comment:

```
R1 NC then R2 APPROVE, same head            -> pass
APPROVE then a second CF says NC, same head -> present_unknown (would be `pass` once P2-1 is fixed)
```

Newest-wins is defensible — an APPROVE after a NEEDS_CHANGES is normally the resolution, and the head pin already discards stale rounds. But once P2-1 lands, two CF seats disagreeing *on the same head* will resolve to `pass` on comment order alone, which is the one remaining path to a false `merge_ready_hint`. Given the PR's own fail-closed theme, prefer: scan all exact-head comments, and let any `needs_changes` win regardless of position.

### P2-3 — the head pin disables itself when `headRefOid` is absent, and does so fail-open

```python
head_short = (head_oid or "")[:7].lower()
...
if head_short:          # empty head_oid -> pinning skipped entirely
```

Exercised through `_enrich_pr` with the key missing from the PR payload:

```
_enrich_pr({no headRefOid}) ->
  {'head': '', 'cf_status': 'pass', 'ci_green': True, 'merge_ready_hint': True}
```

A comment pinning no SHA at all produced `cf_status='pass'` and `merge_ready_hint=True`. `build_default_state` does request `headRefOid` in its `--json` list and `_enrich_pr` has no other call site, so this is unreachable today — but it is the same shape as P1-1 (absent input silently disables the guard rather than refusing), and the file's stated posture is fail-closed. `if not head_oid: return "missing"` at the top.

### P2-4 — still no tests, and the fix pass is exactly what tests are for

Both R1 reviews raised this and it was not a blocker then, so it is not one now. But the five invariants this commit adds — `_repick_seat → None`, unknown CI not green, head-pinned CF, `choice_is_uncertain` fail-closed, `cf_agent=null` propagating to `escalate` — are all silent-regression-prone, and P2-1 through P2-3 above were again found with an ad-hoc driver rather than by CI. The cases are cheap: all-exclude `_repick_seat`, `COMPLETED` with no conclusion, stale-head CF, missing-confidence Choice, and the two real comment bodies from this PR as fixtures.

### Nits

- `ruff format --check` now reports 3 hunks needing reformat (`jev_epic_triage.py:242`, `:486`, `typesafe_client.py:29`) — it was clean at the R1 head, so the fix pass introduced it. `ruff check` is still clean and no CI job gates format, so this is purely local hygiene: one `ruff format` run.
- The `--jq` was changed to `[.[] | {body, createdAt}]`, but the GitHub REST comments payload uses `created_at`, so `createdAt` is always `null`. The field is never read, so nothing breaks — drop it or fix the key.
- When `adj_cf is None`, `escalate` also gains `cf_agent uncertain (top_prob=0.9 < 0.6)` because `escalate_cf` is reused for both causes. The message asserts an inequality that is false (0.9 is not < 0.6). The accurate reason (`no CF seat outside excluded families [...]`) is present alongside it, so the lead is not misled about *whether* to escalate, only about why. Same family of problem as P1-1, much smaller blast radius.
- `_enrich_pr` and `_cf_status` still make two calls to the same comments endpoint per PR (24 round-trips at `--limit 12`).
- PR body test plan is still three unchecked boxes. I verified all three independently and they pass.

---

### Verified

- [x] Exact head: `gh pr view 2756 --json headRefOid` → `6af48f0af9c72b1f44687bcf253216b50b480892` ✅
- [x] CI on that SHA: 11/11 `success`, zero failures, zero pending ✅
- [x] `.venv/bin/ruff check` on both changed files → `All checks passed!`
- [x] `ruff format --check` → 2 files would be reformatted (nit above)
- [x] Behavioural re-test of all four R1 blockers via a driver importing `scripts/jev_epic_triage.py` and `scripts/typesafe_client.py` from a worktree at `6af48f0` — output quoted inline
- [x] Real-comment fixture test of `_cf_status` against the live comment bodies on #2756 and #2758
- [x] Skill mirror parity: `.claude/skills/typesafe-ai/{SKILL,BEST_PRACTICES,COOKBOOKS}.md` byte-identical to `agents_extensions/shared/skills/typesafe-ai/*` — consistent with "Extensions in sync" passing
- [x] Secrets: `KEY_CANDIDATES` is now a single absolute `~/.secrets/typesafe-ai.key`; no cwd-relative path, no key in receipts
- [x] `src/content/` touched: 0 files
- [x] Scope: 8 files, one concern, no generated artifacts
- [ ] Live TypeSafe HTTP call — not executed (no reason to spend quota to review client shape)

### Out of scope

- Live System One API schema, `jev-latest` resolution, and `docs.typesafe.ai` cookbook page bodies — **UNVERIFIABLE** from this seat; not asserted real or fake
- Curriculum modules, workflows, `src/content`

---

### Why APPROVE rather than another round

The bar I published at R1 was P1-1 (fallback refuses and escalates, override string stops claiming exclusion ran) and P1-2 (unknown CI reads as pending, CF head-pinned). Both are met and I confirmed them by execution, not by reading the diff. The P2s above are real and I would like them in a follow-up, but three of the four are fail-closed and the fourth is unreachable on the only call path — none of them is the "reports success while violating the invariant" failure that made R1 a block. Both scripts remain net-new with no call sites, so nothing regresses on merge.

Recommend a follow-up issue covering P2-1 through P2-4 before any lead routes dispatch decisions on `cf_status`, since the gate cannot currently read this repo's own CF comments.

*Cross-family review per `docs/review-protocol.md`. Posted as a comment rather than `gh pr review` because author and reviewer share a GitHub identity.*

